PR #23, opened 2019-05-28
MODINVOICE-35 - Prevent modifications to invoice/invoice lines once status is "Approved" (Java approach)

khandramai:
[MODINVOICE-35](https://issues.folio.org/browse/MODINVOICE-35) - Prevent modifications to invoice/invoice lines once status is "Approved"

## Purpose
When an invoice transitions to "Approved" many of the fields in the invoice and associated invoice lines should not be allowed to be updated.

## Approach
Java-based approach - updating PUT invoice and invoice line endpoints to prevent objects modification.

## Pre-Merge Checklist:


khandramai:
@folio-org/acquisitions ,
according to this story there are two possible ways for implementation. It will be nice if Team provide feedbacks about current java-based approach and the same but Drools-based [#24](https://github.com/folio-org/mod-invoice/pull/24).
Thanks,
Slava

src/main/java/org/folio/rest/impl/InvoicesImpl.java
```diff
@@ -76,11 +85,33 @@ public void putInvoiceInvoicesById(String id, String lang, Invoice invoice, Map<
...
+    invoiceHelper
+      .getInvoice(id)
+      .thenAccept(existed -> {
+        final Consumer<Void> success = ok -> asyncResultHandler.handle(succeededFuture(invoiceHelper.buildNoContentResponse()));
+        if(invoice.getStatus() == Invoice.Status.APPROVED || invoice.getStatus() == Invoice.Status.PAID || invoice.getStatus() == Invoice.Status.CANCELLED) {
+          Set<String> fields = InvoiceProtectedFields.getFieldNames().stream().filter(field -> {
+            try {
+              return !EqualsBuilder.reflectionEquals(FieldUtils.readDeclaredField(invoice, field, true), FieldUtils.readDeclaredField(existed, field, true), true, Invoice.class, true);
+            } catch (IllegalAccessException e) {
+              asyncResultHandler.handle(succeededFuture(invoiceHelper.buildErrorResponse(HttpStatus.HTTP_INTERNAL_SERVER_ERROR.toInt())));
```
piotr-kalashuk: In case of `IllegalAccessException` the `asyncResultHandler` will be completed i.e. the response will be sent back to client but the logic proceeds further

src/main/java/org/folio/rest/impl/InvoicesImpl.java
```diff
@@ -1,28 +1,38 @@
 package org.folio.rest.impl;
 
-import static io.vertx.core.Future.succeededFuture;
```
piotr-kalashuk: Please keep Eclipse like formatting, please, please....

src/main/java/org/folio/rest/impl/InvoicesImpl.java
piotr-kalashuk: The logic here is performed in `thenAccept` block so if any exception happens `exceptionally` should handle it. Please just re-through the exception as `CompitionException` and let `exceptionally` to send response to client

PR #24, opened 2019-05-28
MODINVOICE-35 - Prevent modifications to invoice/invoice lines once status is "Approved" (Drools approach)

khandramai:
[MODINVOICE-35](https://issues.folio.org/browse/MODINVOICE-35) - Prevent modifications to invoice/invoice lines once status is "Approved"

## Purpose
When an invoice transitions to "Approved" many of the fields in the invoice and associated invoice lines should not be allowed to be updated.

## Approach
Drools-based approach - updating PUT invoice and invoice line endpoints to prevent objects modification.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/invoices/utils/ErrorCodes.java
+++ b/src/main/java/org/folio/invoices/utils/ErrorCodes.java
@@ -4,7 +4,8 @@ import org.folio.rest.jaxrs.model.Error;
 
 public enum ErrorCodes {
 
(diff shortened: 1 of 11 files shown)


pautau-s:
I think that in this case the Java approach is preferable, because the Drools approach does not bring any benefits compared to the Java approach, and at the same time has a number of drawbacks:
- it is necessary to learn the new syntax
- creating a new KieSession on every PUT request can affect performance
- you have to take care of releasing KieSession resources

PR #29, opened 2019-05-30
MODINVOICE-50- Calculate Invoice Line totals

KVupp:
## Purpose
https://issues.folio.org/browse/MODINVOICE-50

## Approach
1) Calculate adjustments while GET,POST and PUT for invoice Lines
2) Using JavaMoney API for calculations

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -92,8 +88,9 @@
   }
 
   public CompletableFuture<Void> updateInvoiceLine(InvoiceLine invoiceLine) {
-    return handlePutRequest(resourceByIdPath(INVOICE_LINES, invoiceLine.getId()),
-      JsonObject.mapFrom(invoiceLine), httpClient, ctx, okapiHeaders, logger);
+    return calculateInvoiceLineTotals(invoiceLine)
```
piotr-kalashuk: According to jira description:
1. > Whenever an invoiceLine is created, **updated**, or retrieved, we need to (re)calculate `invoiceLine.adjustmentsTotal` and `invoiceLine.total`.

   Current implementation is just to return `204` status code without body e.g. [PUT /invoice/invoice-lines/{id}](https://s3.amazonaws.com/foliodocs/api/mod-invoice/invoice.html#invoice_invoice_lines__id__put). This is default response type for put responses defined by [item-collection.raml](https://github.com/folio-org/raml/blob/raml1.0/rtypes/item-collection.raml#L63). I discussed this with @craigmcnally and we agreed that at this point no need to change this behavior 
2. > These totals are calculated "on the fly" and never persisted.

   The calculation is done before request is sent to storage which means not readonly properties will be persisted as well

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -122,16 +119,55 @@
...
+    invoiceLine.setId(UUID.randomUUID()
+      .toString());
     invoiceLine.setInvoiceId(invoice.getId());
 
     JsonObject line = mapFrom(invoiceLine);
 
-    return generateLineNumber(invoice)
-      .thenAccept(lineNumber -> line.put(INVOICE_LINE_NUMBER, lineNumber))
+    return generateLineNumber(invoice).thenAccept(lineNumber -> line.put(INVOICE_LINE_NUMBER, lineNumber))
+      .thenAccept(t -> calculateInvoiceLineTotals(invoiceLine, invoice))
```
piotr-kalashuk: According to jira description:
> These totals are calculated "on the fly" and never persisted.

The calculation is done before request is sent to storage which means not readonly properties will be persisted as well

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -122,16 +119,55 @@
...
+    return invoiceLine.getAdjustments()
+      .stream()
+      .map(adj -> calculateAdjustment(adj, subTotal))
+      .collect(MonetaryFunctions.summarizingMonetary(subTotal.getCurrency()))
+      .getSum()
+      .with(MonetaryOperators.rounding());
+  }
+
+
+  private MonetaryAmount calculateAdjustment(Adjustment adjustment, MonetaryAmount subTotal) {
```
piotr-kalashuk: @craigmcnally @KVupp  Just one thing I would like to discuss. There is `relationToTotal` property and according to [Acquisitions Interface Fields - Invoice](https://docs.google.com/spreadsheets/d/1F88PoSnEKD7oFnza1ZuOgzAgKnK95XOyTQzxyww3tZA/edit#gid=131223376&range=A67) 
> In addition to: added to the subtotal, Included in: reported as a portion of the subtotal, Separate from: Calculated based on the subtotal and reported, but not included in the invoice total.

How this property should affect totals?

craigmcnally: That description is somewhat misleading...  

- Adjustments that are "In addition to" the total will affect the total.  
- "Included in" will not affect the total - This cost is baked into the product price - we want to call it out to be explicit about the breakdown of costs but don't need to add the amount to the total.
- "Separate from" will also not affect the total.  - We're going to pay this amount (e.g. maybe to some third party) but it's not actually part of what we're paying the vendor..,

craigmcnally: So basically "Included in" and "Separate from" are solely for bookkeeping.  "In addition" is the only one we need to worry about when calculating the total.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
craigmcnally: So given that we've decided to not return the body on PUT (only 204 as Piotr mentions), there's no need to do the calculation on update... only on POST and GET.  Sorry for the confusion guys.

craigmcnally: Whether the values are persisted or not isn't really all that important - what's important is that the values are calculated upon record retrieval.  If a persisted value is overwritten, so be it.  

It might be slightly clearer if we weren't persisting any values, but since the storage modules are only intended for internal consumption it shouldn't really matter from a practical perspective

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
KVupp: In this case, the values are not persisted, if you look at the **createInvoiceLineSummary** method, the record in storage is being created from the "**line**" object and not the **invoiceLine** object . the calculated values are only returned in the response.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
KVupp: The only reason I added the calculations before storing is to handle some exception scenarios, like invalid currency. In case of update as we do not return anything, we could skip this step or just perform separately as needed.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
KVupp: so should we add the logic into the calculation now, or do we have a separate story for it in the backlog?

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
piotr-kalashuk: okay, got it

pom.xml
```diff
@@ -64,6 +64,11 @@
       <version>${rest-assured.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.javamoney</groupId>
+      <artifactId>moneta</artifactId>
+      <version>1.1</version>
```
craigmcnally: I think 1.3 is available...

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
craigmcnally: it's up to you... I'm fine with either way

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
KVupp: Removed the calculations in PUT for consistency

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
KVupp: ok if a story is not already present, I am handling it in this story

pom.xml
KVupp: Done

PR #36, opened 2019-06-17
MODINVOICE-34  Create voucher on transition to "Approved"

pautau-s:
## Purpose
[MODINVOICE-34](tps://issues.folio.org/browse/MODINVOICE-34) When an invoice transitions to "Approved" status, a voucher and associated voucher lines should be generated

## Approach
Transition triggers if the current invoice status is "Reviewed" or "Open".
* Look for invoice lines by `invoiceId`. 
  * if none is found, the process stops with an error.
* Try to find voucher by `invoiceId`
  * if exist
    * get `systemCurrency` from mod-config
      * if config is empty than use default `systemCurrency`
      * if an error is returned from the mod-config, stop transition to "Approved" with error
    * update extracted voucher object with actual data
  * else
    * get `systemCurrency` from mod-configuration (same logic as if the voucher exists)
    * generate `voucherNumber`
    * build the voucher object on the basis of the invoice.
* Based on the invoice lines found and the voucher built, prepare the data necessary for the generation of voucher lines.
  * group the `List<FundDistribution>` by `fundId`
  * fetch Funds by ids
  * group the `List<Fund>` by `externalAccountNo`
  * map `List<FundDistribution>` to the corresponding `externalAccountNo`.
* Build voucher lines objects on the basis of `Map<externalAccountNo, List<FundDistribution>>`, invoice lines and voucher
* Calculate and set `voucher.amount`
* Handle voucherCreate or update voucher in voucher-storage
  * if  voucher exist
    * update voucher record in voucher-storage
    * look for voucherLines by `voucherId`
      * if lines are found, then delete them
  * else
    * create new record in voucher-storage
* Сreate voucherLines records in storage

Moved all calculation methods to HelperUtils

#### TODOS and Open Questions
It is not clear where to get the values for the fields: "accountingCode", "exportToAccounting", "type", "status". Therefore, for now I set the hard-coded default values for these fields in `InvoiceHelper.setDefaultRequiredFields()` method.
- [ ] Start using real data to create a voucher when it becomes known where to get these values from.

## Pre-Merge Checklist:


pom.xml
```diff
@@ -14,15 +14,18 @@
   </licenses>
 
   <properties>
+    <jsonschema_paths>acq-models/mod-invoice-storage/schemas,raml-util/schemas,acq-models/common/schemas/</jsonschema_paths>
+    <raml-module-builder.version>24.0.0</raml-module-builder.version>
```
piotr-kalashuk: Duplicated/conflicting lines

src/main/java/org/folio/invoices/rest/exceptions/HttpException.java
```diff
@@ -16,6 +16,11 @@ public HttpException(int code, String message) {
     this.error = new Error().withCode(ErrorCodes.GENERIC_ERROR_CODE.getCode()).withMessage(message);
   }
 
+  public HttpException(int code, ErrorCodes errorCode) {
+    this.code = code;
+    this.error = new Error().withCode(errorCode.getCode()).withMessage(errorCode.getDescription());
```
piotr-kalashuk: `ErrorCodes` has `toError()` which does same thing

pautau-s: Done

pom.xml
pautau-s: Removed

src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -68,7 +81,7 @@ private HelperUtils() {
 
   public static CompletableFuture<JsonObject> getVoucherById(String id, String lang, HttpClientInterface httpClient, Context ctx,
       Map<String, String> okapiHeaders, Logger logger) {
-    String endpoint = String.format(GET_VOUCHER_BY_ID, id, lang);
+    String endpoint = String.format(VOUCHER_BY_ID_ENDPOINT, id, lang);
```
KVupp: permission to get voucher by id is missing "voucher-storage.vouchers.item.get" in module descriptor

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -147,15 +178,287 @@
...
+    }
+    return VertxCompletableFuture.completedFuture(null);
+  }
+
+  private boolean isTransitionToApproved(Invoice invoiceFromStorage, Invoice invoice) {
+    return invoiceFromStorage.getStatus() == Invoice.Status.REVIEWED && invoice.getStatus() == Invoice.Status.APPROVED;
+  }
+
+  /**
+   * Handles transition of given invoice to Approved status.
```
KVupp: Maintainability: It's really helpful how the logic is detailed in the PR, if it is present here in the comments section, it would be helpful in the longer run

src/main/java/org/folio/invoices/utils/HelperUtils.java
pautau-s: This permission is not used when transitioning to "Approved". The getVouchers() method is used to obtain an existing voucher. 
Renamed this constant to make it more generic because it is used not only for the GET, but also for PUT.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Added doc comments

craigmcnally:
> It is not clear where to get the values for the fields: "accountingCode", "exportToAccounting", "type", "status". Therefore, for now I set the hard-coded default values for these fields in InvoiceHelper.setDefaultRequiredFields() method.

* `accountingCode` (currently `erpCode`) comes from the vendor record:  https://github.com/folio-org/acq-models/blob/163af732853b1a7a4cc8e9b48f57f7f7cd7cb387/mod-orgs/schemas/organization.json#L100  I believe this will eventually be renamed to `accountingCode` but a story doesn't not yet exist.
* `exportToAccounting`:  From the interface fields document:  
> This would keep the invoice from being feed into the batch process (Not generate a external voucher/payment) but would still move values in the system. Note: this is ideally defined by the vendor relationship and exposed for override on the invoice. (Not yet added to Vendor Information)
* `type`:  for now just use "Voucher".  I'll confirm with the POs

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+
+  private Error buildFundNotFoundError(List<String> idsNotFound) {
+    Parameter parameter = new Parameter().withKey("fundIds").withValue(idsNotFound.toString());
+    return new Error()
+      .withCode(FUNDS_NOT_FOUND.getCode())
+      .withMessage(FUNDS_NOT_FOUND.getDescription())
+      .withParameters(Collections.singletonList(parameter));
+  }
+
+  private CompletableFuture<List<Fund>> getFoundsByIds(List<String> ids) {
```
craigmcnally: Typo... `Founds` -> `Funds`

src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -231,4 +242,83 @@ private static boolean isFieldNotEmpty(Object newObject, Object existedObject, S
...
+   * @return String representing CQL query to get records by id's
+   */
+  public static String convertIdsToCqlQuery(List<String> ids) {
+    return StreamEx.of(ids).map(id -> "id==" + id).joining(" or ");
+  }
+
+  /**
+   * Wait for all requests completion and collect all resulting objects. In case any failed, complete resulting future with the exception
+   * @param futures list of futures and each produces resulting object on completion
+   * @param <T> resulting type
```
javalkar-varun: This param does not exist. Can be removed.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+  /**
+   * Prepares a new voucher or updates existing one for further processing
+   *
+   * @param invoice {@link Invoice} to be approved on the basis of which the voucher is prepared
+   * @return completable future with {@link Voucher} on success
+   */
+  private CompletableFuture<Voucher> prepareVoucher(Invoice invoice) {
+    return findExistingVoucher(invoice.getId())
+      .thenCompose(vouchers -> {
+        if (isNotEmpty(vouchers)){
```
craigmcnally: whitespace before `{`

src/main/java/org/folio/rest/impl/AbstractHelper.java
```diff
@@ -57,6 +65,45 @@
     setDefaultHeaders();
   }
 
+  /**
+   * Retrieve configuration by moduleName and configName from mod-configuration.
+   *
+   * @param moduleName name of the module for which the configuration is to be retrieved
+   * @param configName name of the configuration to retrieve it
+   * @return CompletableFuture with JsonObject
```
javalkar-varun: This should be `CompletableFuture with List of Config objects` ?

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+  /**
+   * Build new {@link Voucher} based on processed {@link Invoice}
+   *
+   * @param invoice invoice {@link Invoice} to be approved
+   * @return completable future with {@link Voucher}
+   */
+  private CompletableFuture<Voucher> buildNewVoucher(Invoice invoice) {
+
+    Voucher voucher = new Voucher();
+    voucher.setInvoiceId(invoice.getId());
```
craigmcnally: For now this is fine - we'll eventually have to revisit this once we implement aggregate and batch vouchers which relate to multiple invoices.

src/test/java/org/folio/rest/impl/ApiTestBase.java
```diff
@@ -127,12 +128,12 @@ Response verifyPostResponse(String url, String body, Headers headers, String exp
   }
 
   Response verifyPut(String url, String body, String expectedContentType, int expectedCode) {
-    Headers headers = prepareHeaders(X_OKAPI_URL, X_OKAPI_TENANT);
+    Headers headers = prepareHeaders(X_OKAPI_URL, X_OKAPI_TENANT,X_OKAPI_TOKEN);
```
javalkar-varun: Missing space after comma `, `

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+   */
+  private CompletableFuture<Void> approveInvoice(Invoice invoice) {
+    CompletableFuture<Void> future = new VertxCompletableFuture<>(ctx);
+    getInvoiceLinesWithTotals(invoice)
+      .thenApply(invoiceLines -> {
+        verifyInvoiceLineNotEmpty(invoiceLines);
+        return invoiceLines;
+      })
+      .thenCombine(prepareVoucher(invoice), (invoiceLines, voucher) -> {
+        invoice.setVoucherNumber(voucher.getVoucherNumber());
```
craigmcnally: didn't notice that we have this two-way relationship... Making a note to see if we can avoid this.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+      });
+  }
+
+  /**
+   * Prepares the data necessary for the generation of voucher lines based on the invoice lines found
+   *
+   * @param invoiceLines {@link List<InvoiceLine>} associated with processed {@link Invoice}
+   * @return {@link InvoiceLine#fundDistributions} grouped by {@link Fund#externalAccountNo}
+   */
+  private CompletableFuture<Map<String, List<FundDistribution>>> groupFundDistributionsByExternalAccountNo(
```
craigmcnally: Wow that's a long method name!  Not a huge deal but might make sense to abbreviate... 

`groupFundDistroByExternalAccountNo` or even `groupFundDistroByExternalAcctNo`

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+      List<InvoiceLine> invoiceLines) {
+
+    Map<String, List<FundDistribution>> fundDistributionsGroupedByFundId = getGroupFundDistributionsByFundId(invoiceLines);
```
craigmcnally: again... `Distributions` -> `Distros` is slightly shorter but it's up to you.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+      .map(this::getFoundsByIds)
+      .collect(toList());
+
+    return collectResultsOnSuccess(futures)
+      .thenApply(results -> results
+        .stream()
+        .flatMap(List::stream)
+        .collect(toList())
+      )
+      .thenApply(existedFunds -> verifyThatAllFundsFound(existedFunds, fundIds));
```
craigmcnally: nitpick...  using the present tense makes more sense IMO `existingFunds` instead of `existedFunds`

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+  }
+
+  private CompletableFuture<List<Fund>> getFoundsByIds(List<String> ids) {
+    String query = encodeQuery(HelperUtils.convertIdsToCqlQuery(ids), logger);
+    String endpoint = String.format(FUND_SEARCHING_ENDPOINT, query, ids.size(), lang);
+    return handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger)
+      .thenApply(this::extractFounds);
+  }
+
+  private List<Fund> extractFounds(JsonObject entries) {
```
craigmcnally: again... `Founds`

pautau-s:
> * `accountingCode` (currently `erpCode`) comes from the vendor record:  https://github.com/folio-org/acq-models/blob/163af732853b1a7a4cc8e9b48f57f7f7cd7cb387/mod-orgs/schemas/organization.json#L100  I believe this will eventually be renamed to `accountingCode` but a story doesn't not yet exist.
> * `exportToAccounting`:  From the interface fields document:
> 
> > This would keep the invoice from being feed into the batch process (Not generate a external voucher/payment) but would still move values in the system. Note: this is ideally defined by the vendor relationship and exposed for override on the invoice. (Not yet added to Vendor Information)

Does this work need to be done in the scope of the current task?

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +176,355 @@
...
+  }
+
+  /**
+   * Gets exchange rate based on {@link Voucher#invoiceCurrency} and {@link Voucher#systemCurrency} using default ExchangeRateProvider
+   *
+   * @param voucher {@link Voucher} for which the exchange rate factor is obtained
+   * @return {@link Double} value of exchange rate factor
+   */
+  private CompletableFuture<Double> getExchangeRate(Voucher voucher) {
+    return VertxCompletableFuture.supplyAsync(() -> MonetaryConversions.getExchangeRateProvider()
```
craigmcnally: It would be nice if you could get and stash the `CurrencyConversion` object and then pass it to `calculateVoucherLineAmount(...)` instead of the voucher.  That would make the conversion code much cleaner IMO.

craigmcnally: Ah, looks like you either need the voucher anyway, or to pass the invoiceCurrency into that method.

src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -231,4 +242,83 @@ private static boolean isFieldNotEmpty(Object newObject, Object existedObject, S
...
+    MonetaryAmount voucherLineAmount = Money.of(0, invoiceCurrency);
+
+    for (FundDistribution fundDistribution : fundDistributions) {
+      InvoiceLine invoiceLine = findLineById(invoiceLines, fundDistribution.getInvoiceLineId());
+      MonetaryAmount subTotal = Money.of(invoiceLine.getTotal(), invoiceCurrency);
+
+      voucherLineAmount = voucherLineAmount.add(subTotal.with(MonetaryOperators.percent(fundDistribution.getPercentage())));
+    }
+
+    MonetaryAmount convertedAmount = voucherLineAmount
```
craigmcnally: Per my earlier comment(s) this could be simplified to `voucherLine.with(conversion)` if we had the `CurrencyConversion` object...

src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -231,4 +242,83 @@ private static boolean isFieldNotEmpty(Object newObject, Object existedObject, S
...
+  public static double calculateVoucherAmount(Voucher voucher, List<VoucherLine> voucherLines) {
+
+    CurrencyUnit currency = Monetary.getCurrency(voucher.getSystemCurrency());
+
+    MonetaryAmount amount = voucherLines.stream()
+      .map(line -> Money.of(line.getAmount(), currency))
+      .collect(MonetaryFunctions.summarizingMonetary(currency))
+      .getSum();
+
+    return convertToDouble(amount);
```
craigmcnally: I realize that convertToDouble wasn't added as part of this PR, but we probably want to add a javadoc comment or rename the method to indicate that it's performing some rounding.  It's easy to screw up calculations if you start rounding in places where you shouldn't, so we want to make sure it's evident that we're not just converting to double here.

craigmcnally:
> > * `accountingCode` (currently `erpCode`) comes from the vendor record:  https://github.com/folio-org/acq-models/blob/163af732853b1a7a4cc8e9b48f57f7f7cd7cb387/mod-orgs/schemas/organization.json#L100  I believe this will eventually be renamed to `accountingCode` but a story doesn't not yet exist.
> > * `exportToAccounting`:  From the interface fields document:
> > 
> > > This would keep the invoice from being feed into the batch process (Not generate a external voucher/payment) but would still move values in the system. Note: this is ideally defined by the vendor relationship and exposed for override on the invoice. (Not yet added to Vendor Information)
> 
> Does this work need to be done in the scope of the current task?

No.  The defaults you provided are adequate - Thanks for checking, sorry I wasn't clear

src/main/java/org/folio/invoices/utils/HelperUtils.java
pautau-s: I initially intended to use CurrencyConversion, but then I realized that voucherLineAmount.with(conversion) is a blocking operation and so I've chosen the hard way.

src/main/java/org/folio/invoices/utils/HelperUtils.java
pautau-s: Renamed

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Done

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Done

src/main/java/org/folio/rest/impl/AbstractHelper.java
pautau-s: Fixed

src/test/java/org/folio/rest/impl/ApiTestBase.java
pautau-s: Done

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: The voucherLines also refers twice to invoiceLines: 
- [sourceIds](https://github.com/folio-org/acq-models/blob/163af732853b1a7a4cc8e9b48f57f7f7cd7cb387/mod-invoice-storage/schemas/voucher_line.json#L28)
- [fundDistributions[].invoiceLineId](https://github.com/folio-org/acq-models/blob/163af732853b1a7a4cc8e9b48f57f7f7cd7cb387/mod-invoice-storage/schemas/fund_distribution.json#L20)

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Shortened it.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Done

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: corrected

src/main/java/org/folio/invoices/utils/HelperUtils.java
pautau-s: https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#@param
Tried to make the doc comment more clear.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
craigmcnally: Thanks for pointing this out

src/main/java/org/folio/invoices/utils/HelperUtils.java
craigmcnally: I don't think it is... The conversion is obtained from the provider earlier - this is probably the more expensive operation which could block.  Apply the conversion itself should be fairly fast - it's probably doing essentially the same thing you're doing here...

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +177,340 @@
...
+        if (isNotEmpty(vouchers)) {
+          Voucher voucher = vouchers.get(0);
+          return updateExistingVoucherState(invoice, voucher);
         }
+        return buildNewVoucher(invoice);
       });
   }
 
+  private CompletableFuture<List<Voucher>> findExistingVoucher(String invoiceId) {
+    return voucherHelper.getVouchers(1, 0, String.format(QUERY_BY_INVOICE_ID, invoiceId))
```
piotr-kalashuk: There is `getVoucherByInvoiceId(String invoiceId)` method in `VoucherHelper` which might be used instead

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -149,15 +177,340 @@
...
+    voucher.setInvoiceCurrency(invoice.getCurrency());
+    setDefaultRequiredFields(voucher);
+
+    return VertxCompletableFuture.completedFuture(voucher);
+  }
+
+  //TODO Start using real information to create a voucher when it becomes known where to get these values from.
+  private void setDefaultRequiredFields(Voucher voucher) {
+    voucher.setAccountingCode(DEFAULT_ACCOUNTING_CODE);
+    voucher.setExportToAccounting(false);
```
piotr-kalashuk: I believe you can use latest version of `acq-models` to get `exportToAccounting` value from `invoice` as per [Acquisitions Interface Fields - Voucher](https://docs.google.com/spreadsheets/d/1F88PoSnEKD7oFnza1ZuOgzAgKnK95XOyTQzxyww3tZA/edit#gid=1784784202&range=K7)
> Export this payment record to a financial system if connected properly. Setting inherited from invoice

pautau-s: Thank you for pointing me.

src/main/java/org/folio/rest/impl/InvoiceHelper.java
pautau-s: Removed

src/main/java/org/folio/invoices/utils/HelperUtils.java
pautau-s: Yeah, you're right. Simplified to `voucherLineAmount.with(conversion)`.

src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -284,20 +286,16 @@ public static double calculateVoucherAmount(Voucher voucher, List<VoucherLine> v
   public static Double calculateVoucherLineAmount(List<FundDistribution> fundDistributions, List<InvoiceLine> invoiceLines, Voucher voucher) {
     CurrencyUnit invoiceCurrency = Monetary.getCurrency(voucher.getInvoiceCurrency());
     CurrencyUnit systemCurrency = Monetary.getCurrency(voucher.getSystemCurrency());
+    CurrencyConversion conversion = MonetaryConversions.getConversion(systemCurrency, IMF_EXCHANGE_RATE_PROVIDER);
```
craigmcnally: Sorry to be a pain... I don't think I was clear earlier.  We probably want to avoid "getting" the conversion for each line for two reasons:

1) You could potentially wind up with different conversion rates across your voucher lines if the rate happens to update in the middle of voucher generation (unlikely, but I think possible)
2) Performance.

In `InvoiceHelper.setExchangeRateFactor(...)` you could get the conversion like you do here, get the exchange rate from it and save it so you can later pass it in to this method.  That way you know that the voucher and all the voucher lines have the same converstion/exchangeRate, and you're only incurring the cost of looking up the exchange rate from the provider once.

PR #58, opened 2019-08-13
MODINVOICE-87 Calculate and persist invoice totals upon invoice line updates

piotr-kalashuk:
## Purpose
[MODINVOICE-87](https://issues.folio.org/browse/MODINVOICE-87) Calculate and persist invoice totals
[MODINVOICE-86](https://issues.folio.org/browse/MODINVOICE-86) Calculate and persist invoice line totals on creation

## Approach
[MODINVOICE-87](https://issues.folio.org/browse/MODINVOICE-87):
- Trigger calculations only if invoice is still open for modifications i.e. status is `OPEN` or `REVIEWED`
- When invoice line is retrieved and totals are outdated, the record is persisted in storage and on success triggers an event to update invoice
- Adding `adjustmentsTotal`, `subTotal` and `total` as protected i.e. cannot be modified anymore after invoice approval

[MODINVOICE-86](https://issues.folio.org/browse/MODINVOICE-86):
- Updating logic to store invoice line totals in storage
- Adding unit tests to check that `POST` invoice line record in response is the same as one sent to storage

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/InvoiceHelper.java
+++ b/src/main/java/org/folio/rest/impl/InvoiceHelper.java
@@ -28,13 +28,14 @@ import static org.folio.invoices.utils.HelperUtils.getInvoiceById;
 import static org.folio.invoices.utils.HelperUtils.handleDeleteRequest;
 import static org.folio.invoices.utils.HelperUtils.handleGetRequest;
 import static org.folio.invoices.utils.HelperUtils.handlePutRequest;
-import static org.folio.invoices.utils.HelperUtils.isFieldsVerificationNeeded;
+import static org.folio.invoices.utils.HelperUtils.isPostApproval;
 import static org.folio.invoices.utils.ResourcePathResolver.FOLIO_INVOICE_NUMBER;
 import static org.folio.invoices.utils.ResourcePathResolver.FUNDS;
 import static org.folio.invoices.utils.ResourcePathResolver.INVOICES;
 import static org.folio.invoices.utils.ResourcePathResolver.ORDER_LINES;
 import static org.folio.invoices.utils.ResourcePathResolver.resourceByIdPath;
 import static org.folio.invoices.utils.ResourcePathResolver.resourcesPath;
+import static me.escoffier.vertx.completablefuture.VertxCompletableFuture.completedFuture;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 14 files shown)


src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -156,20 +172,29 @@ private void updateOutOfSyncInvoiceLine(InvoiceLine invoiceLine) {
...
+      .thenCompose(invoiceLineFromStorage -> getInvoice(invoiceLineFromStorage).thenCompose(invoice -> {
+        // 3. Validate if invoice line update is allowed
+        validateInvoiceLine(invoice, invoiceLine, invoiceLineFromStorage);
+        invoiceLine.setInvoiceLineNumber(invoiceLineFromStorage.getInvoiceLineNumber());
+
+        // 4. Recalculate totals before update which also indicates if invoice requires update
+        calculateInvoiceLineTotals(invoiceLine, invoice);
+
+        // 5. Update invoice line in storage
+        return updateInvoiceLineToStorage(invoiceLine).thenRun(() -> {
```
javalkar-varun: Will invoiceLine be updated to storage before `calculateInvoiceLineTotals` even completes?

piotr-kalashuk: `calculateInvoiceLineTotals` returns `InvoiceLine` which may be updated with actual totals. This is blocking operation i.e. `updateInvoiceLineToStorage` can start only when `calculateInvoiceLineTotals` completed its work

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -241,6 +248,14 @@ private void updateOutOfSyncInvoice(Invoice invoice) {
     });
   }
 
+  private CompletableFuture<Boolean> recalculateTotals(Invoice updatedInvoice, Invoice invoiceFromStorage) {
```
grabsefx: As for method returns boolean value, probably it is better to name it something like `isTotalsRecalculationRequired(..)` for better code readability.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -103,24 +105,37 @@ private void updateOutOfSyncInvoiceLine(InvoiceLine invoiceLine) {
...
    * Calculate invoice line total and compare with original value if it has changed
    *
-   * @param invoiceLineFromStorage invoice line to update totals for
-   * @return {code true} if grand total value is different to original one
+   * @param invoiceLine invoice line to update totals for
+   * @param invoice invoice record
+   * @return {code true} if any total value is different to original one
    */
-  public CompletableFuture<Boolean> reCalculateInvoiceLineTotals(InvoiceLine invoiceLineFromStorage) {
+  private boolean reCalculateInvoiceLineTotals(InvoiceLine invoiceLine, Invoice invoice) {
```
grabsefx: method name `isInvoiceLineTotalsRecalculationRequired` is more meaningful from my opinion

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -208,19 +231,13 @@ private Invoice checkIfInvoiceLineCreationAllowed(Invoice invoice) {
    * @return completable future which might hold {@link InvoiceLine} on success or an exception if any issue happens
    */
   CompletableFuture<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine, Invoice invoice) {
-    invoiceLine.setId(UUID.randomUUID()
-      .toString());
+    invoiceLine.setId(randomUUID().toString());
```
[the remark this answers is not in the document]
piotr-kalashuk: Agree, removed it

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
javalkar-varun: I do not see any error messages in the console with respect to this operation so I think it should be fine. We have handled it asynchronously in other places so that’s why I was curious https://github.com/folio-org/mod-invoice/blob/master/src/main/java/org/folio/rest/impl/InvoiceLineHelper.java#L137

PR #65, opened 2019-08-20
MODINVOICE-54 Prorated adjustments - handle invoice update

piotr-kalashuk:
## Purpose
[MODINVOICE-54](https://issues.folio.org/browse/MODINVOICE-54) Prorated adjustments part 1 - basics

## Approach
Apply prorated adjustments to each invoice line when invoice is updated:
1. Get invoice from storage
1. Compare prorated adjustments in the updated invoice and the one from storage
   - If nothing changed, just proceed with the logic not related to prorated adjustments
   - If there is difference, get all the invoice lines of this invoice
1. The updated invoice either has new adjustment (a), existing one has different content (b) or some adjustment is no longer available (c).
   a. Add new adjustment to each invoice line
   b. Update corresponding adjustment in each line
   c. Remove the adjustment from each line
1. Recalculate invoice line totals (adjustment total and grand total)
1. Persist each invoice line
1. Proceed updating the invoice (basically recalculate totals and persist invoice)

The cases when invoice already has some prorated adjustments and user does something with invoice lines:
1. Create new line
   - Get invoice
   - Get already existing invoice lines
   - Redistribute prorated adjustments
   - Persist new line
   - Persist updates for existing lines
   - Persist new invoice's totals
1. Update existing line changing subTotal or quantity
   - Get invoice. Check if there is prorated adjustment affected by the change
   - Get already existing invoice lines
   - Reapply prorated adjustments
   - Persist updates for all existing lines
   - Persist new invoice's totals
1. Delete line
   - Get invoice line (to get invoice id)
   - Delete line
   - Get invoice
   - Get other existing invoice lines
   - Redistribute prorated adjustments
   - Persist updates for existing lines
   - Persist new invoice's totals

### TODOS and Open Questions
- [x] If grand `subTotal` is zero it is unclear how to apply prorated adjustment by amount. Example:
  - Invoice is created to cover gifts. Each invoice line has zero sub total. User tries to add "delivery fee" and intentionally (or by mistake) adds it as prorated adjustment "By amount"
  - Invoice has a few lines with positive and negative `subTotal`s which compensates each other. For example, the case when some resources are returned back to vendor (e.g. manufacturer issues) and new are ordered instead
- [x] Which scale to use when distributing percentage adjustment "By lines" e.g. `5/3=1.(6)`. Current implementation uses scale = 2
- [x] What should happen if user adds negative prorated adjustment to an invoice which has mixed invoice lines (positive and negative sub totals)?
- [ ] For example, lets consider invoice with 3 lines and user updates invoice adding prorated adjustment. At step 5 two lines are persisted successfully but update of the third one failed for some reason.
  - Should we proceed with invoice update or return an error to client (or probably the invoice should be updated first and the lines afterwards)?
  - What to do with the lines which persisted successfully ?
  - Does the logic should reattempt to persist failed lines ?
At the moment generic error will be return and invoice is not changed in storage but updated lines remain as is i.e. no any roll-back
- [ ] There are quite a lot of places where something might go wrong when dealing with invoice lines and the content of the lines (user actually did not touch explicitly) might become incorrect in case of any error.
- [ ] Also what cases can be handled asynchronously and which must be done as part of operation on invoice line before sending reply to client...

## Pre-Merge Checklist:


dennisbridges:
If grand subTotal is zero it is unclear how to apply prorated adjustment by amount. Example:
Invoice is created to cover gifts. Each invoice line has zero sub total. User tries to add "delivery fee" and intentionally (or by mistake) adds it as prorated adjustment "By amount"
Invoice has a few lines with positive and negative subTotals which compensates each other. For example, the case when some resources are returned back to vendor (e.g. manufacturer issues) and new are ordered instead **DB: If all zero they would all have the same amount applied. If negative they would be prorated as if positive amounts. (-(-x))**
  Which scale to use when distributing percentage adjustment "By lines" e.g. 5/3=1.(6). Current implementation uses scale = 2
  What should happen if user adds negative prorated adjustment to an invoice which has mixed invoice lines (positive and negative sub totals)? **DB: These would be applied as is, the negative would lower the positive total and increase the negative total. I don't know of a use case for this happening so it makes sense to me not to work any magic with it**

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -276,19 +308,244 @@ private String getInvoiceLineNumberEndpoint(String id) {
...
+   */
+  private void applyProratedAdjustmentByLines(Adjustment adjustment, List<InvoiceLine> lines, CurrencyUnit currencyUnit) {
+    Adjustment proratedAdjustment = prepareAdjustmentForLine(adjustment);
+
+    if (adjustment.getType() == Adjustment.Type.AMOUNT) {
+      proratedAdjustment.setValue(convertToDoubleWithRounding(Money.of(adjustment.getValue(), currencyUnit)
+        .divide(lines.size())));
+    } else {
+      proratedAdjustment.setValue(BigDecimal.valueOf(adjustment.getValue())
+        .divide(BigDecimal.valueOf(lines.size()), 2, RoundingMode.HALF_EVEN)
```
craigmcnally: I thought we were using JavaMoney API for all these calculations... this looks like you're just using BigDecimal.  I'm not sure it really matters, but I want to make sure we're not rounding and losing precision unnecessarily.  The scaling factor of 2 and RoundingMode caught my attention

craigmcnally: Now I understand your question about scale... 🤔

craigmcnally: I implemented this part differently in my POC so need to work though it - I'll try to get you an answer tonight

piotr-kalashuk: `BigDecimal` is used here because we are not working with money yet but just splitting percentage value across all the lines. Once the adjustment's value is calculated, it will add money portion of this particular adjustment to `line.adjustmentsTotal`

piotr-kalashuk: I do not think that there might be any issue with performance. For example, `5d/3d=1.6666666666666667` and we can store this as is. The only thing is related to UI whether to display such values as is or applying some trimming...

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -279,31 +279,91 @@ private void updateOutOfSyncInvoice(Invoice invoice) {
...
+      }
+
+      // Collecting original hash codes of each invoice line record.
+      List<Integer> originalLineHashCodes = calculateHashCodes(lines);
+
+      invoiceLineHelper.processProratedAdjustments(lines, updatedInvoice);
+
+      return VertxCompletableFuture.allOf(ctx, lines.stream()
+        // If any update is made to invoice line adjustment(s), the new hash code will be generated for invoice line record
+        .filter(line -> !originalLineHashCodes.contains(line.hashCode()))
```
pautau-s: There is no guarantee that after the line change the new hash code will not match the hash code of another line on the list. Since this is related to monetary calculations, I would suggest using a more reliable filtering method.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -276,19 +308,244 @@ private String getInvoiceLineNumberEndpoint(String id) {
...
+      helper.updateInvoiceAndLines(invoiceId).handle((ok, fail) -> {
+        helper.closeHttpClient();
+        return null;
+      });
+    });
+  }
+
+  private CompletableFuture<Void> updateInvoiceAndLines(String invoiceId) {
+    return getInvoiceById(invoiceId, lang, httpClient, ctx, okapiHeaders, logger)
+      .thenCompose(invoice -> getInvoiceLinesByInvoiceId(invoiceId).thenApply(lines -> {
```
pautau-s: As with the `applyProratedAdjustments(InvoiceLine invoiceLine, Invoice invoice)` method, there is no need to retrieve and recalculate the lines if there are no prorated adjustments in the invoice.

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
piotr-kalashuk: The change applied to not use any scale in this case

src/main/java/org/folio/rest/impl/InvoiceHelper.java
piotr-kalashuk: Changes applied to use array list which holds invoice lines affected by prorated adjustment and may contain duplicates. When final result is returned, distinct values are only taken into account

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
piotr-kalashuk: Added extra condition

PR #91, opened 2019-12-18
MODINVOICE-121 - Release 3.0.1

khandramai:
MODINVOICE-121 - Release 3.0.1

### DIFF
--- a/ramls/invoice.raml
+++ b/ramls/invoice.raml
@@ -1,7 +1,7 @@
 #%RAML 1.0
 
 title: Invoice
-version: v3.0
+version: v3.1
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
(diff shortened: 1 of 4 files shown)


ramls/invoice.raml
```diff
@@ -1,7 +1,7 @@
 #%RAML 1.0
 
 title: Invoice
-version: v3.0
+version: v3.1
```
KVupp: is this version increment necessary? We haven't modified anything in the request/response

khandramai: In scope of MODINVOICE-120 hardcoded voucher "accounting code" was changed to real accounting code from invoice. This affects voucher creation during invoice approval flow, i.e invoice updating. Is this not a reason for minor version increasing?

KVupp: yea that is the reason to increase the interface version . But what does it have to do with the raml version?

khandramai: Yes, correct. I've already reverted. Thanks.

PR #102, opened 2020-02-27
[Modinvoice-130] - Create batch-voucher API

AndreiMPV:
[MODINVOICE-130](https://issues.folio.org/browse/MODINVOICE-130) - Create batch-voucher API

## Purpose
Define and implement the batch-voucher API.
API is able to retrieve batch voucher as XML and JSON based on Accept header.

## Approach
1. JAXB Maven plugin was introduce for generation models from XSD schemes.
2. Wrappers classes for supporting JAXB were created:
- **org.folio.config.JAXBConfig** -> Spring configuration. Contains all necessary JAXB beans.
- **org.folio.jaxb.JAXBContextWrapper** -> Facade thread-safe class for creation Unmarshaller and Marshaller.
- **org.folio.jaxb.JAXBUtil** -> Contains utility methods to help deal with JAXB models
- **org.folio.jaxb.XMLConverter** -> Responsible for marshaling and unmarshaling JAXB models to XML responces.

## Pre-Merge Checklist:


craigmcnally:
I was surprised to see you use spring framework here - this is something that we haven't done before in any of our other modules.  What is the rationale for pulling this in?

AndreiMPV:
> I was surprised to see you use spring framework here - this is something that we haven't done before in any of our other modules. What is the rationale for pulling this in?

Spring always be in the Folio and ThunderJet team used it before in some places. The main reason to avoid creation service beans by using **new**. Also there is a Spring implementation for ConversationService.

craigmcnally:
@AndreiMPV my concern is consistency - the use of spring in and of itself isn't a problem, but it's not used elsewhere in the module, let alone in other acquisition modules (AFAIK).  

The dependency injection stuff is convenient, and if we were writing a new module I'd advocate its use.  However, I'm not crazy about the inconsistency it adds to a somewhat mature module.  

The ConversionService OTOH is something new to me, but seems straightforward, and might be a more compelling reason for pulling in spring.

In the end, I can't really justify asking you to rewrite this.  That's not the best use of your time and I'd feel like a real jerk doing so.  However, I will ask that you bring it up at a standup next time you want to pull in a new dependency that could lead to significant differences in new code from existing code.  Want to pull in an Apache-commons?  Fine, go right ahead, NBD.  Want to pull in something like spring that introduces new annotations, etc. just mention it during standup - "Hey guys, I'm thinking about using this thing because it provides X, Y, and Z.  What do you think?"

src/main/java/org/folio/jaxb/JAXBUtil.java
```diff
@@ -0,0 +1,87 @@
...
+      throw new ConversionFailedException(Instant.class, XMLGregorianCalendar.class, e);
+    }
+    return result;
+  }
+
+  private static Class<?> loadClass(String className) {
+    try {
+      return Class.forName(className);
+    } catch (ClassNotFoundException e) {
+      return null;
```
khandramai: Maybe it makes sense to throw exception here instead of return null from block catch?

craigmcnally: I tend to agree - if a class can't be found here, isn't it likely to result in something blowing up later in the processing?  I'd think it would be better to surface the issue ASAP.  If we throw an exception from here, we might as well just remove this method altogether and simply use `Class.forName(...)` directly.

src/main/java/org/folio/jaxb/JAXBDefaultNamespacePrefixMapper.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Implementation of {@link NamespacePrefixMapper} that maps the schema
+ * namespaces more to readable names. Used by the jaxb marshaller. Requires
+ * setting the property "com.sun.xml.bind.namespacePrefixMapper" to an instance
+ * of this class.
+ * <p>
+ * Requires dependency on JAXB implementation jars
+ * </p>
+ */
+@SuppressWarnings("squid:S1191") //The com.sun.xml.bind.marshaller.NamespacePrefixMapper is part of jaxb logic
```
craigmcnally: See https://issues.apache.org/jira/browse/SIS-126

src/main/java/org/folio/rest/impl/BatchVoucherHelper.java
```diff
@@ -0,0 +1,94 @@
...
+        return null;
+      });
+    return future;
+  }
+
+  private Response buildSuccessResponse(JsonObject jsonObjectBatchVoucher, String acceptHeader) {
+    BatchVoucher jsonBatchVoucher = jsonObjectBatchVoucher.mapTo(BatchVoucher.class);
+    switch (acceptHeader) {
+    case APPLICATION_XML:
+      BatchVoucherType xmlBatchVoucher = conversionService.convert(jsonBatchVoucher, BatchVoucherType.class);
```
khandramai: @AndreiMPV @craigmcnally ,
do we need such complex solution? I tried to use Jackson here and do the following:
`        BatchVoucherType xmlBatchVoucher = conversionService.convert(jsonBatchVoucher, BatchVoucherType.class);
        String xmlBatchVoucherResponse = xmlConverter.marshal(xmlBatchVoucher, true);

        ObjectMapper objectMapper = new ObjectMapper();
        ObjectMapper xmlMapper = new XmlMapper();
        JsonNode tree = objectMapper.readTree(jsonObjectBatchVoucher.encode());
        String jsonAsXml = xmlMapper.writer().withRootName("batchVoucher").writeValueAsString(tree);
        return buildSuccessResponse(jsonAsXml, APPLICATION_XML);`

and all the unit tests passed successfully.

khandramai: @AndreiMPV @craigmcnally ,
looks like this solution is very complex& I just tried to use Jackson here:
`    ObjectMapper objectMapper = new ObjectMapper();
        ObjectMapper xmlMapper = new XmlMapper();
        JsonNode tree = objectMapper.readTree(jsonObjectBatchVoucher.encode());
        String jsonAsXml = xmlMapper.writer().withRootName("atchVoucher").writeValueAsString(tree);
        return buildSuccessResponse(jsonAsXml, APPLICATION_XML);`
instead of
`BatchVoucherType xmlBatchVoucher = conversionService.convert(jsonBatchVoucher, BatchVoucherType.class);
        String xmlBatchVoucherResponse = xmlConverter.marshal(xmlBatchVoucher, true);`
and unit tests passed successfully.

src/main/java/org/folio/jaxb/JAXBDefaultNamespacePrefixMapper.java
AndreiMPV: Thanks for pointing on solution. 
NameSpacePrefixMapper was deleted

AndreiMPV:
> @AndreiMPV my concern is consistency - the use of spring in and of itself isn't a problem, but it's not used elsewhere in the module, let alone in other acquisition modules (AFAIK).
> 
> The dependency injection stuff is convenient, and if we were writing a new module I'd advocate its use. However, I'm not crazy about the inconsistency it adds to a somewhat mature module.
> 
> The ConversionService OTOH is something new to me, but seems straightforward, and might be a more compelling reason for pulling in spring.
> 
> In the end, I can't really justify asking you to rewrite this. That's not the best use of your time and I'd feel like a real jerk doing so. However, I will ask that you bring it up at a standup next time you want to pull in a new dependency that could lead to significant differences in new code from existing code. Want to pull in an Apache-commons? Fine, go right ahead, NBD. Want to pull in something like spring that introduces new annotations, etc. just mention it during standup - "Hey guys, I'm thinking about using this thing because it provides X, Y, and Z. What do you think?"

Hi @craigmcnally,

With rest of the team we decided delete Spring DI and I've initialize all services manually.

src/main/java/org/folio/rest/impl/BatchVoucherHelper.java
AndreiMPV: @khandramai @craigmcnally
Thanks for pointing. 
I agree with you and also not so happy with a bunch of Converters and codes around that.
There are alternative approaches how to get XML from response:
1. I also tried to implement this functionality using the method that Slava advised, but failed to achieve the correct collection conversion (List<BatchedVoucher>) and also in the conversion of DateTime types.
2. There was a other solution, which basen  on **org.folio.rest.tools.plugins.ResourceMethodExtensionPlugin**, which able to update Java sources and add annotations to existing models. In this case, we could use the scheme only for validation. Unfortunately, I was not able to implement this approach, some problems  were occurred in the implementation.
3. At the moment, it was partially possible, using reflection, to annotate the JSON models with JAXB anotations and this makes it possible to eliminate the needed to generate separate models according to the XSD scheme and at the same time use the validation of the produced XML according to the XSD scheme.
With such a decision, we can 
a) still use JAXBMarshaller for the production of XML(OOTB used schema validation)
b) or use Jackson XmlMapperr
c) or use JAXB utility class.
For me the best to use JAXBMarshaller.
Thanks,
Andrei

AndreiMPV:
> You have 2 major code smells and 2 minor. Please address at least the major ones, if not all 4.

Resolved

src/main/java/org/folio/rest/impl/BatchVoucherHelper.java
craigmcnally: We use JaxbMarshaller in both mod-gobi and mod-oai-pmh, so there's some precedence there.  I would say that schema validation is strongly desired, if not a hard requirement.  If that's not possible (or possible but difficult) with the other approaches the decision is clear.

That said, I'm not sure I fully understand the need to use reflection - we didn't run into that in either of the aforementioned modules.  What's different here that necessitates this extra complexity?

craigmcnally: Also note that there's precedence for using jackson in edge-rtac.  Here we created the POJOs ourselves with the necessary annotations.  IIRC we generated the POJOs from the schema once, copied them into the main codebase and augmented them with necessary annotations/code.

craigmcnally: BTW... https://github.com/folio-org/mod-oai-pmh/blob/795d0c72b898e43688f26da152225b02ff0d91bd/src/main/java/org/folio/oaipmh/ResponseHelper.java#L30

D'oh!

khandramai:
@craigmcnally ,

khandramai:
@craigmcnally ,
looks like current problem is different than we resolved in oai-pmh, trace, etc. Here we need to transform json to xml with xsd-validation. I briefly reviewed Jackson and wasn't able to find solution for this problem (namely, for xsd validation). This approach works despite looks not so easy. So I can propose to merge it as is now but we can create spike for investigation of this problem in more detail. This allows us to delivery this functionality to release. What do you think about this?

khandramai:
@craigmcnally ,
like an option we can use xsd-schema in raml to specify type of response. This allows us to prevent additional conversion that Andrei made and use only Jackson. We reported this problem for oai-pmh but looks it isn't resolved yet https://issues.folio.org/browse/RMB-270.

AndreiMPV:
> You have 2 major code smells and 2 minor. Please address at least the major ones, if not all 4.

Resolved.

PR #272, opened 2021-07-21
MODSOURMAN-527 - Update dependency on processing-core library

RuslanLavrov:
## Purpose
it is necessary to update dependency on data-import-processing-core library to pick up fix for EDIFACT records data retrieving

## Learning
[MODSOURMAN-527](https://issues.folio.org/browse/MODSOURMAN-527)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,7 +159,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>data-import-processing-core</artifactId>
-      <version>3.1.1</version>
+      <version>3.2.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.mockftpserver</groupId>


Aliaksandr-Fedasiuk (approved):

grabsefx:
Could we wait for release version of this dependency, not the snapshot?

KaterynaSenchenko:
Hi @grabsefx, we need to test the changes before the release (which might happen today), but to do so on hosted ref envs - we need to update the dependency on snapshot version

KaterynaSenchenko (approved):

RuslanLavrov:
Hello @grabsefx, it is necessary to verify the fix on the hosted envs and to unblock the library release. We plan to release the library in the nearest future. As soon as the library is released, I will update the dependency on the release version.

grabsefx (approved):

grabsefx:
> Hi @grabsefx, we need to test the changes before the release (which might happen today), but to do so on hosted ref envs - we need to update the dependency on snapshot version

Actualy, you dont need to create PR for this purpose. You could just push your changes to the separate brunch, jenkins will build the docker container for you, and you can deploy it on any reference environment manually

PR #343, opened 2022-02-21
[MODINVOICE-360] Unrelease encumbrances when cancelling invoice

damien-git:
## Purpose
[MODINVOICE-360](https://issues.folio.org/browse/MODINVOICE-360) - Unrelease auto-released encumbrances when cancelling an invoice

## Approach
See Jira ticket.

#### TODOS and Open Questions
[MODFIN-236](https://issues.folio.org/browse/MODFIN-236) - Missing permissions for payment/credit PUT

## Learning
Implementing this would have been easier with [MODFISTO-268](https://issues.folio.org/browse/MODFISTO-268) because the transaction summary id would not need to be an order id or an invoice id. Even though we are cancelling an invoice, we have to use an order transaction summary to modify the encumbrances. Also this implementation will have to be updated after we implement MODFISTO-268.
There is an issue when using Mockito to stub methods returning a null value: because the default mock returns a null value, we cannot know if the given stub was invoked or not. The solution is to define a Mockito default answer for the mock, and return an exception instead of null. Together with `verifyNoMoreInteractions()`, this ensures that all the stubs are called and only the ones declared. This solution does not work with Mock annotations, so we have to use `Mockito.mock()`.

## Pre-Merge Checklist:

  - [ ] Duplications on new code is 3% or less
  
### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -532,10 +532,6 @@
       "version": "3.0"
     },
     {
-      "id": "orders",
-      "version": "12.0"
-    },
-    {
       "id": "voucher-storage.vouchers",
       "version": "1.2"
     },
@@ -614,6 +610,12 @@
     {
       "id": "data-import-converter-storage",
       "version": "1.2"
+    }
+  ],
+  "optional": [
+    {
+      "id": "orders",
+      "version": "12.0"
     }
   ],
   "permissionSets": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -54,6 +54,8 @@
             "invoice-storage.invoices.item.get",
             "invoice-storage.invoice-lines.item.put",
             "invoice-storage.invoice-lines.collection.get",
+            "orders.collection.get",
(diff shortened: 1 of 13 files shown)


damien-git:
Screenshot_20220221_092215
Screenshot_20220221_154750
failing tests were failing before

AndreiMPV (approved):

Dmitriy-Butramyou (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -527,6 +531,10 @@
       "id": "order-lines",
       "version": "3.0"
     },
+    {
```
AndreiMPV, line 534: Looks like will be circulation dependency between mod-invoice and mod-orders

AndreiMPV (changes requested):

descriptors/ModuleDescriptor-template.json
damien-git: Thanks. I am not sure there would be a circular dependency because invoice is optional in mod-orders. But I think you are right anyway that orders should be optional in mod-invoice. I moved that to optional.

AndreiMPV (approved):

PR #371, opened 2022-08-01
[Modinvoice 417] Provide endpoint to validate fund distributions

SinghAdes:
## Purpose

https://issues.folio.org/browse/MODINVOICE-417
## Approach

Provide new endpoint to validate fund distributions  for a Invoice line & adjustments that will be called by UI to have consistent validation with BE
## Pre-Merge Checklist:


descriptors/ModuleDescriptor-template.json
```diff
@@ -667,6 +674,11 @@
       "displayName": "Invoice Line - get collection of Invoice lines",
       "description": "Get collection of Invoice lines"
     },
+    {
+      "permissionName": "invoice.invoice-lines.fund-distributions.validate",
+      "displayName": "Orders - validate fund distributions for a invoice line",
```
SerhiiNosko, line 679: This should be Invoices

src/main/java/org/folio/rest/impl/InvoicesImpl.java
```diff
@@ -174,6 +188,40 @@ public void deleteInvoiceInvoiceLinesById(String id, String lang, Map<String, St
...
+  public void putInvoiceInvoiceLinesFundDistributionsValidate(ValidateFundDistributionsRequest entity,
+     Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+      Context vertxContext) {
+    try {
+      MonetaryAmount subTotal;
+      Double total = null;
+      CurrencyUnit currencyUnit = Monetary.getCurrency(entity.getCurrency());
+      List<FundDistribution> fundDistributionList = entity.getFundDistribution();
+      if(CollectionUtils.isNotEmpty(entity.getAdjustments()))
+      {
```
SerhiiNosko, line 201: Proper format this code

src/main/java/org/folio/rest/impl/InvoicesImpl.java
```diff
@@ -174,6 +188,40 @@ public void deleteInvoiceInvoiceLinesById(String id, String lang, Map<String, St
       .exceptionally(t -> handleErrorResponse(asyncResultHandler, invoiceLineHelper, t));
   }
 
+  @Override
+  public void putInvoiceInvoiceLinesFundDistributionsValidate(ValidateFundDistributionsRequest entity,
```
SerhiiNosko, line 192: Better to rename entity to request

src/main/java/org/folio/rest/impl/InvoicesImpl.java
```diff
@@ -174,6 +188,40 @@ public void deleteInvoiceInvoiceLinesById(String id, String lang, Map<String, St
...
+      CurrencyUnit currencyUnit = Monetary.getCurrency(entity.getCurrency());
+      List<FundDistribution> fundDistributionList = entity.getFundDistribution();
+      if(CollectionUtils.isNotEmpty(entity.getAdjustments()))
+      {
+      if (InvoiceValidator.isAdjustmentIdsNotUnique(entity.getAdjustments())) {
+       throw new HttpException(400,ADJUSTMENT_IDS_NOT_UNIQUE);
+      }
+      subTotal = Money.of(entity.getSubTotal(), currencyUnit);
+      MonetaryAmount adjustmentAndFundTotals = HelperUtils.calculateAdjustmentsTotal(entity.getAdjustments(), subTotal);
+      total = HelperUtils.convertToDoubleWithRounding(adjustmentAndFundTotals.add(subTotal));
```
SerhiiNosko, line 207: You can declare total variable here (in place of usage)

src/main/java/org/folio/services/adjusment/AdjustmentsService.java
```diff
@@ -48,7 +48,7 @@ public List<Adjustment> getProratedAdjustments(InvoiceLine invoiceLine) {
     return filterAdjustments(invoiceLine.getAdjustments(), INVOICE_LINE_PRORATED_ADJUSTMENT_PREDICATE);
   }
 
-  private List<Adjustment> filterAdjustments(List<Adjustment> adjustments, Predicate<Adjustment> predicate) {
+  public static List<Adjustment> filterAdjustments(List<Adjustment> adjustments, Predicate<Adjustment> predicate) {
```
SerhiiNosko, line 51: AdjustmentsService is not utility class, so it is a bad practice to make method as static here.
You can create instance of this service using constructor

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -85,7 +92,7 @@ public void validateInvoiceTotals(Invoice invoice) {
     }
   }
 
-  private boolean isAdjustmentIdsNotUnique(List<Adjustment> adjustments) {
+  public static boolean isAdjustmentIdsNotUnique(List<Adjustment> adjustments) {
```
SerhiiNosko, line 95: The same here, you don't need static here, you can use this validator as in other places using constructor: new InvoiceValidator();

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -98,7 +105,7 @@ public void validateBeforeApproval(Invoice invoice, List<InvoiceLine> lines) {
     validateInvoiceTotals(invoice);
     verifyInvoiceLineNotEmpty(lines);
     validateInvoiceLineFundDistributions(lines, Monetary.getCurrency(invoice.getCurrency()));
-    validateInvoiceAdjustmentsDistributions(adjustmentsService.getNotProratedAdjustments(invoice) , Monetary.getCurrency(invoice.getCurrency()));
+    validateInvoiceAdjustmentsDistributions(adjustmentsService.getNotProratedAdjustments(invoice),Monetary.getCurrency(invoice.getCurrency()));
```
SerhiiNosko, line 108: I see couple of places with incorrect formatting of commas, please format it. The correct formatting: comma + space + next operand

src/test/java/org/folio/rest/impl/InvoiceLinesApiTest.java
```diff
@@ -280,6 +286,37 @@ public void deleteInvoiceLinesBadLanguageTest() {
     verifyDeleteResponse(endpoint, TEXT_PLAIN, 400);
   }
 
+  @Test
+  public void fundValidationTest()
+  {
```
SerhiiNosko, line 291: The same with brackets formatting, use the same formatting as in other methods

src/test/resources/mockdata/fundsValidator/validateFundDistributionsRequest.json
```diff
@@ -0,0 +1,53 @@
+{
+  "subTotal": 10,
+  "currency": "USD",
+  "adjustments": [
```
SerhiiNosko, line 4: you can format json using this tool: https://jsonformatter.curiousconcept.com/#

src/test/java/org/folio/rest/impl/InvoiceLinesApiTest.java
```diff
@@ -280,6 +286,37 @@ public void deleteInvoiceLinesBadLanguageTest() {
...
+    ValidateFundDistributionsRequest reqData = getMockAsJson(FUND_VALIDATOR_MOCK_DATA_PATH).mapTo(ValidateFundDistributionsRequest.class);
+     verifyPut(INVOICE_LINE_FUNDS_VALIDATOR_ID_PATH, reqData, "", 204);
+  }
+  @Test
+  public void fundValidationDupAdjIdTest()
+  {
+    ValidateFundDistributionsRequest reqData = getMockAsJson(FUND_VALIDATOR_MOCK_DATA_PATH_DUP_ADJ_ID).mapTo(ValidateFundDistributionsRequest.class);
+     verifyPut(INVOICE_LINE_FUNDS_VALIDATOR_ID_PATH, reqData, "", 400);
```
SerhiiNosko, line 299: You also need to check remainingAmount field in your test in some way, not only response codes, because this field is very important to test calculation logic

src/main/java/org/folio/exceptions/ExceptionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+      errors = new Errors().withErrors(Collections.singletonList(GENERIC_ERROR_CODE.toError()
+                           .withAdditionalProperty(ERROR_CAUSE, cause.getMessage())))
+                           .withTotalRecords(1);
+    }
+    return errors;
+  }
+
+  @SuppressWarnings("java:S5852")
+  public static boolean isErrorMessageJson(String errorMessage) {
+    if (!StringUtils.isEmpty(errorMessage)) {
```
Abdulkhakimov, line 41: You can use StringUtils.isNotEmpty static method

okolawole-ebsco:
Filling out the purpose and approach of a PR is a good way to communicate context to reviewers. An alternative is if a link to the JIRA ticket modified is good as well. Can you please add one or both of the options?

src/main/java/org/folio/exceptions/ExceptionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      errors = new Errors().withErrors(Collections.singletonList(GENERIC_ERROR_CODE.toError()
+                           .withAdditionalProperty(ERROR_CAUSE, cause.getMessage())))
+                           .withTotalRecords(1);
+    }
+    return errors;
+  }
+
+  @SuppressWarnings("java:S5852")
+  public static boolean isErrorMessageJson(String errorMessage) {
```
okolawole-ebsco, line 40: please add a comment here describing what the method does, example input, etc.

src/main/java/org/folio/exceptions/ExceptionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+      Matcher matcher = pattern.matcher(errorMessage);
+      if (matcher.find()) {
+        return matcher.groupCount() == 3;
+      }
+    }
+    return false;
+  }
+
+
+  private static Error mapToError(Error error) {
```
okolawole-ebsco, line 52: I would expect a `mapToError` method will be mapping from another POJO to an `Error`, this doesn't seem to be the case. Please add a comment to this method describing what it is supposed to do.

src/main/java/org/folio/rest/impl/BaseApi.java
```diff
@@ -85,27 +85,37 @@ protected HttpException handleProcessingError(Throwable throwable) {
   }
 
   public Response buildErrorResponse(Throwable throwable) {
-    return buildErrorResponse(handleProcessingError(throwable));
+  logger.error("Exception encountered", throwable.getCause());
```
okolawole-ebsco, line 88: The indentation on this line seems to be off by a few spaces.  Please correct this.

src/main/java/org/folio/rest/impl/BaseApi.java
```diff
@@ -85,27 +85,37 @@ protected HttpException handleProcessingError(Throwable throwable) {
...
+    final Response.ResponseBuilder responseBuilder = createResponseBuilder(code);
+    return responseBuilder.header(CONTENT_TYPE, APPLICATION_JSON)
+      .entity(errors)
+      .build();
   }
 
-  public Response buildErrorResponse(HttpException exception) {
-    final Response.ResponseBuilder responseBuilder;
-    switch (exception.getCode()) {
+  public static int defineErrorCode(Throwable throwable) {
```
okolawole-ebsco, line 97: nitpick: this method could return `javax.ws.rs.core.Response.Status` rather than its status code. The switch statement in `createResponseBuilder` method below would operate on `javax.ws.rs.core.Response.Status` rather than `int`. To clarify, this method would return `INTERNAL_SERVER_ERROR` object instead of its error code.

The code smell became more obvious because the the default return in the switch statement still relies on `javax.ws.rs.core.Response.Status`.

src/main/java/org/folio/rest/impl/InvoicesImpl.java
okolawole-ebsco, line 192: By just eyeballing this method, I see no indication that any processing is deferred(async with Future/callbacks). If this method calls out to external parties or the database, it will be good to relinquish the vertx event loop. My concern is that this method would hog the event loop if the duration of this method is not quick. This method is fine as is, if there arent't any long operations contained.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
       if (CollectionUtils.isEmpty(adjustment.getFundDistributions())) {
         throw new HttpException(400, ADJUSTMENT_FUND_DISTRIBUTIONS_NOT_PRESENT);
       }
-
-      if (isFundDistributionSummaryNotValid(adjustment, currencyUnit)) {
-        throw new HttpException(400, ADJUSTMENT_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH);
+      validateFundDistributionForInvoiceLine(adjustment.getValue(), adjustment.getFundDistributions(), currencyUnit);
+    }
+  }
+  public static void validateFundDistributionForInvoiceLine(Double total, List<FundDistribution> fundDistributions,CurrencyUnit currencyUnit) {
```
okolawole-ebsco, line 142: the name of this method implies that it is meant to validate fund distributions for an invoice line, but its parameters do not accepts and invoice line. There ought to be representation in the parameter list that represents the invoice line.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+    Double subtotal = Money.of(total,currencyUnit).getNumber().doubleValue();
```
okolawole-ebsco, line 143: for my education, can you describe to me the essence of this line?

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
-    return ObjectUtils.notEqual(sumMixedDistributions(line, currencyUnit), line.getTotal());
-  }
 
-  private Double sumMixedDistributions(InvoiceLine line, CurrencyUnit currencyUnit) {
-    return sumMixedDistributions(line.getFundDistributions(), line.getTotal(), currencyUnit);
-  }
-
-  private void validateInvoiceAdjustmentsDistributions(List<Adjustment> adjustments, CurrencyUnit currencyUnit) {
-    for (Adjustment adjustment : adjustments){
+  public static void validateInvoiceAdjustmentsDistributions(List<Adjustment> adjustments,CurrencyUnit currencyUnit) {
```
okolawole-ebsco, line 134: Looks like this method is validating adjustments AND distributions but i didnt get that this method name. It seems a distribution is a child/component of the adjustment so i don't think it needs to be reflected in the method name. Adding a comment to the method is a good way to show that the adjustment and the distributions underneath are validated.

An example that illustrates what I am mean; `validateUsersFirstNameLastName`. To validate the user means to validate the first name and the last name and possibly other criteria. First name and last name dont need to be in the method name. `validateUsers` would suffice instead.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+  }
+  public static void validateFundDistributionForInvoiceLine(Double total, List<FundDistribution> fundDistributions,CurrencyUnit currencyUnit) {
+    Double subtotal = Money.of(total,currencyUnit).getNumber().doubleValue();
+    if (subtotal != null && CollectionUtils.isNotEmpty(fundDistributions)) {
+      if (subtotal == 0d) {
+        validateZeroPrice(fundDistributions);
+        return;
       }
+      BigDecimal remainingPercent = ONE_HUNDRED_PERCENT;
+      for (FundDistribution fundDistribution : fundDistributions) {
```
okolawole-ebsco, line 150: in general, what do you think is the cardinality of fund distributions? is it in the tens, hundreds, thousands or general see less then 5 here?

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+        FundDistribution.DistributionType dType = fundDistribution.getDistributionType();
+        if (dType == PERCENTAGE) {
+          remainingPercent = remainingPercent.subtract(BigDecimal.valueOf(fundDistribution.getValue()));
+        } else {
+          Double value = fundDistribution.getValue();
```
okolawole-ebsco, line 155: nitpick: you could put this line above the `if` statement and substitute it for any location that `fundDistribution.getValue` is needed.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+        }
+      }
+      checkRemainingPercentMatchesToZero(remainingPercent, subtotal);
     }
   }
-
-  private boolean isFundDistributionSummaryNotValid(Adjustment adjustment, CurrencyUnit currencyUnit) {
-    return ObjectUtils.notEqual(sumMixedDistributions(adjustment.getFundDistributions(), adjustment.getValue(), currencyUnit), adjustment.getValue());
+  public static void validateZeroPrice(List<FundDistribution> fdList) {
+   FundDistribution.DistributionType firstFdType = fdList.get(0).getDistributionType();
```
okolawole-ebsco, line 166: validate that the list is not empty before retrieving the first item.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+    } else {
+      BigDecimal remainingPercent = ONE_HUNDRED_PERCENT;
+      for (FundDistribution fd : fdList
+      ) {
+        remainingPercent = remainingPercent.subtract(BigDecimal.valueOf(fd.getValue()));
+      }
+      checkRemainingPercentMatchesToZero(remainingPercent, 0d);
+    }
+  }
+  private static void checkRemainingPercentMatchesToZero(BigDecimal remainingPercent, Double subtotal) {
```
okolawole-ebsco, line 183: add a empty line above this line.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+      BigDecimal epsilon = BigDecimal.valueOf(1e-10);
+      if (remainingPercent.abs().compareTo(epsilon) > 0) {
+        throwExceptionWithIncorrectAmount(remainingPercent, subtotal);
+      }
+  }
+  private static void throwExceptionWithIncorrectAmount(BigDecimal remainingPercent, Double subtotal) {
```
okolawole-ebsco, line 189: add an empty line above this line.

src/main/java/org/folio/services/validator/InvoiceValidator.java
okolawole-ebsco, line 142: I am seeing a lot of `BigDecimal` conversion in this method and maybe this class in general, can we curb this somehow? Some operations appears to be possible without converting to a `BigDecimal` first.

descriptors/ModuleDescriptor-template.json
SinghAdes: done.

src/main/java/org/folio/rest/impl/InvoicesImpl.java
SinghAdes: done

src/main/java/org/folio/rest/impl/InvoicesImpl.java
SinghAdes: i guess we can not, aswe are using "total" variable outside of the if else scope

src/main/java/org/folio/services/adjusment/AdjustmentsService.java
SinghAdes: yes,i agree.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: done

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: done

src/test/java/org/folio/rest/impl/InvoiceLinesApiTest.java
SinghAdes: done

src/test/resources/mockdata/fundsValidator/validateFundDistributionsRequest.json
SinghAdes: done

src/test/java/org/folio/rest/impl/InvoiceLinesApiTest.java
SinghAdes: sure , updated test

src/test/resources/mockdata/invoiceLines/invoice_lines.json
```diff
@@ -50,7 +50,7 @@
           "relationToTotal": "In addition to"
         }
       ],
-      "adjustmentsTotal": 0.22,
+      "adjustmentsTotal": 0.0,
```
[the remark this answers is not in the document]
SinghAdes: yes i agree,
That was because i was testing with 0 adjustmentsTotal. it should be fixed now. thanks

src/main/java/org/folio/exceptions/ExceptionUtil.java
SinghAdes: done

src/main/java/org/folio/exceptions/ExceptionUtil.java
SinghAdes: i have updated the logic for error  response building and removed this entire class , requesting you to please review

src/main/java/org/folio/exceptions/ExceptionUtil.java
SinghAdes: i have updated the logic for error  response building and removed this entire class , requesting you to please review

src/main/java/org/folio/rest/impl/BaseApi.java
SinghAdes: done

src/main/java/org/folio/rest/impl/BaseApi.java
SinghAdes: i have updated the logic for error  response building and removed this method , requesting you to please review

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: formatting done

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: formatting done

descriptors/ModuleDescriptor-template.json
SerhiiNosko: Maybe I wrote suggestion not using much details, we need to rename this line "Orders - validate fund distributions for a invoice line" to this line "Invoices - validate fund distributions for a invoice line".
We don't need to touch permission names, if all permissions named in singular manner - we need to use the same pattern for new permission as well

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -115,46 +115,82 @@ private void verifyInvoiceLineNotEmpty(List<InvoiceLine> invoiceLines) {
...
-      if (isFundDistributionSummaryNotValid(line, currencyUnit)) {
-        throw new HttpException(400, LINE_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH);
-      }
+      validateFundDistributions(line.getTotal(), line.getFundDistributions(), currencyUnit);
     }
   }
 
-  private boolean isFundDistributionSummaryNotValid(InvoiceLine line, CurrencyUnit currencyUnit) {
-    return ObjectUtils.notEqual(sumMixedDistributions(line, currencyUnit), line.getTotal());
+  public static void validateAdjustments(List<Adjustment> adjustments, CurrencyUnit currencyUnit) {
```
SerhiiNosko, line 126: This method also should not be static

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -115,46 +115,82 @@ private void verifyInvoiceLineNotEmpty(List<InvoiceLine> invoiceLines) {
...
+      if (CollectionUtils.isEmpty(adjustment.getFundDistributions())) {
+        throw new HttpException(400, ADJUSTMENT_FUND_DISTRIBUTIONS_NOT_PRESENT);
+      }
+      validateFundDistributions(adjustment.getValue(), adjustment.getFundDistributions(), currencyUnit);
+    }
   }
 
-  private Double sumMixedDistributions(InvoiceLine line, CurrencyUnit currencyUnit) {
-    return sumMixedDistributions(line.getFundDistributions(), line.getTotal(), currencyUnit);
+  public static void validateFundDistributions(Double total, List<FundDistribution> fundDistributions, CurrencyUnit currencyUnit) {
```
SerhiiNosko, line 135: The same here, please remove static

ramls/invoice.raml
```diff
@@ -176,6 +176,21 @@ resourceTypes:
         description: Update an invoice line with given {id}
       delete:
         description: Delete an invoice line with given {id}
+    /fundDistributions/validate:
```
SerhiiNosko, line 179: In proposed implementation endpoint path would be like this: /invoice-lines/fundDistributions/validate
So there is mixing of styles, we need to use consistent approach in naming, so you need to change path to this one: /invoice-lines/fund-distributions/validate
Also it is good to be consisted with mod-orders where the same naming convention chosen:
https://github.com/folio-org/mod-orders/blob/master/descriptors/ModuleDescriptor-template.json#L311

src/main/java/org/folio/invoices/utils/ErrorCodes.java
```diff
@@ -17,7 +17,9 @@ public enum ErrorCodes {
   FUND_DISTRIBUTIONS_NOT_PRESENT("fundDistributionsNotPresent", "At least one fund distribution should present for every associated invoice line"),
   ACCOUNTING_CODE_NOT_PRESENT("accountingCodeNotPresent", "Invoice can not be approved, because it requires an accounting code to be export to accounting"),
   ADJUSTMENT_FUND_DISTRIBUTIONS_NOT_PRESENT("adjustmentFundDistributionsNotPresent", "At least one fund distribution should present for every non-prorated adjustment"),
-  LINE_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH("lineFundDistributionsSummaryMismatch", "Fund distributions summary should be 100 % or equal to subtotal for every associated invoice lines"),
+  LINE_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH("lineFundDistributionsSummaryMismatch", "Fund distributions summary should be 100 % or equal to subtotal"),
+  CANNOT_MIX_TYPES_FOR_ZERO_PRICE("cannotMixTypesForZeroPrice", "Fund distribution types cannot be mixed for a price of 0"),
+  INCORRECT_FUND_DISTRIBUTION_TOTAL("incorrectFundDistributionTotal","Fund distribution total must add to 100% or totalPrice"),
```
SerhiiNosko, line 22: Constants LINE_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH and INCORRECT_FUND_DISTRIBUTION_TOTAL are equal by meaning, can we use only one?

src/test/resources/mockdata/fundsValidator/validateFundDistributionsRequest_duplicate_adjId.json
```diff
@@ -0,0 +1,53 @@
+{
+  "subTotal": 10,
+  "currency": "USD",
+  "adjustments": [
+      {
```
SerhiiNosko, line 5: This file also should be formatted

src/main/java/org/folio/rest/impl/BaseApi.java
```diff
@@ -90,6 +91,9 @@ public Response buildErrorResponse(Throwable throwable) {
 
   public Response buildErrorResponse(HttpException exception) {
     final Response.ResponseBuilder responseBuilder;
+    Errors errors = exception.getErrors();
+    List<Error> errorList = errors.getErrors().stream().collect(toList());
```
SerhiiNosko, line 95: errors.getErrors().stream().collect(toList()) can be replaced just as: errors.getErrors()

src/test/resources/mockdata/fundsValidator/validateFundDistributionsRequest_duplicate_adjId.json
SinghAdes: it should be fixed now
.

src/main/java/org/folio/invoices/utils/ErrorCodes.java
SinghAdes: its fixed now.

ramls/invoice.raml
SinghAdes: yup, i agree.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: yup , it should be fixed now.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: done.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: we are validating list in caller method & i have modified this method as private now.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: yes i agree, it should be fixed now

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: yes i agree, it should be fixed now

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: well , i guess it would be max in tens.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: it is creating a Money instance with a specific currency and amount without any rounding. The logic works without this line also but as it was existing logic in mod-invoice that's why kept.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: yes i agree, it is fixed now.

src/main/java/org/folio/rest/impl/InvoicesImpl.java
SinghAdes: well it is just for validating fund amounts, i dont think it would contain any long operations.

src/main/java/org/folio/services/validator/InvoiceValidator.java
```diff
@@ -119,42 +126,76 @@ private void validateInvoiceLineFundDistributions(List<InvoiceLine> invoiceLines
...
+          remainingPercent = remainingPercent.subtract(percentageValue);
+        }
+      }
+      checkRemainingPercentMatchesToZero(remainingPercent, subtotal);
     }
   }
-
-  private boolean isFundDistributionSummaryNotValid(Adjustment adjustment, CurrencyUnit currencyUnit) {
-    return ObjectUtils.notEqual(sumMixedDistributions(adjustment.getFundDistributions(), adjustment.getValue(), currencyUnit), adjustment.getValue());
+  public static void validateZeroPrice(List<FundDistribution> fdList) {
```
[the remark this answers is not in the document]
SinghAdes: We are validating "firstFdType" value below so i guess we should not change . i would be thankful if you can guide . thanks in advance.

src/main/java/org/folio/rest/impl/BaseApi.java
SinghAdes: it should be fixed now.

src/main/java/org/folio/services/validator/InvoiceValidator.java
okolawole-ebsco: so the goal is to get an amount that does not have rounding?

I would say if the line doesn't add value, it should be removed.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SerhiiNosko: @okolawole-ebsco basically this logic is very similar to logic that we provided for order lines, and all this logic tested for orders, so I propose to leave as it for now.
Here is validation for orders: https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/orders/utils/FundDistributionUtils.java#L37

okolawole-ebsco: I see. We can leave this for testing sake. My concern is about all the boxing and unboxing going on, which is performance related but I can't tell the impact from looking at the code.

An approach i would have taken is to have the POJO store as BigDecimal instead of Double since BigDecimal is what we want. I am certain, this change might break other things...

src/main/java/org/folio/services/validator/InvoiceValidator.java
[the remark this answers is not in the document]
okolawole-ebsco: This is a nitpick, so you can choose to not adhere to this suggestion. Here is an alternative I have:
```
private void validateZeroPrice(List<FundDistribution> fdList) {
    FundDistribution.DistributionType firstFdType = fdList.get(0).getDistributionType();
    BigDecimal remainingPercent = ONE_HUNDRED_PERCENT;
    for (FundDistribution fd : fdList) {
      if (fd.getDistributionType() != firstFdType) {
        throw new HttpException(422, CANNOT_MIX_TYPES_FOR_ZERO_PRICE);
      }

      if (firstFdType == AMOUNT) {
        if (fd.getValue() != 0) {
          throwExceptionWithIncorrectAmount(ZERO_REMAINING_AMOUNT);
        }
      } else {
        remainingPercent = remainingPercent.subtract(BigDecimal.valueOf(fd.getValue()));
      }
    }
    checkRemainingPercentMatchesToZero(remainingPercent, 0d);
  }
```

src/main/java/org/folio/rest/impl/InvoicesImpl.java
okolawole-ebsco: Sounds good.

src/main/java/org/folio/services/validator/InvoiceValidator.java
SinghAdes: i have removed line.

SerhiiNosko (approved):

AndreiMPV (approved):

PR #405, opened 2023-04-07
[MODINVOICE-462] - Skip POL status update when paying for previous FY

Dmitriy-Butramyou:
## Pre-Merge Checklist:

  - [x] Duplications on new code are 3% or less


Abdulkhakimov (approved):

src/main/java/org/folio/services/invoice/InvoicePaymentService.java
```diff
@@ -49,10 +57,30 @@ public Future<Void> payInvoice(Invoice invoice, List<InvoiceLine> invoiceLines,
...
+    String fundID = invoiceLines.stream()
+      .flatMap(invoiceLine -> invoiceLine.getFundDistributions().stream())
+      .map(FundDistribution::getFundId)
+      .findFirst()
+      .orElse(null);
+
+    if (StringUtils.isNotBlank(fundID)) {
+      return currentFiscalYearService.getCurrentFiscalYearByFund(fundID, requestContext)
+        .compose(currentFiscalYear -> {
+          if (Objects.equals(currentFiscalYear.getId(), invoice.getFiscalYearId())) {
```
damien-git, line 74: I think `invoice.getFiscalYearId()` can be `null`, in which case the previous behavior should be applied (with no additional call).

Dmitriy-Butramyou: I think we should change this field to required after changing the UI.

grabsefx: > I think we should change this field to required after changing the UI.

in that case we should also create migration script which will populate this field for existing invoices. And I am not sure if it is possible

damien-git: Right, the fiscal year information would be in mod-finance-storage as a transaction, assuming the invoice lines are linked to order lines (which is not even guaranteed), or if we just want to use the current fiscal year it would also be in a mod-finance-storage table. We would not be able to use that information from a migration script in mod-invoice. So we should handle the case of a null fiscalYearId. This will also make it simpler for the UI code.

PR #413, opened 2023-05-25
[MODINVOICE-478] Prevent conflicts when saving multiple invoice lines with data import

damien-git:
## Purpose
[MODINVOICE-478](https://issues.folio.org/browse/MODINVOICE-478) - Data import saving invoice lines in parallel can cause conflicts when invoices are updated

## Approach
New implementation to save invoice and invoice lines with data import.
There will no longer be potential conflicts when the invoice is updated, which happens several times in the process.
Line creations are processed together, which should make the whole operation significantly faster.
Error reporting is a little different: some (internal) errors that were previously reported with a line number will no longer be associated with a number. Other errors (for validation) will be better reported and will prevent the creation of the invoice and lines.

#### TODOS and Open Questions
TODO: Integration tests, especially `data-import/src/main/resources/folijet/data-import/features/import-edi-invoice.feature`. I was not able to test it because I could not get Kafka messages to be sent to IDEA from mod-data-import.
Also, we should check that error reporting still matches user expectations.

## Learning
I learned a way to validate objects with their Json schema. We should do that every time we don't get the objects directly with the API (such as when using Kafka).

## Pre-Merge Checklist:

  - [ ] Duplications on new code are 3% or less

### DIFF
--- a/src/main/java/org/folio/services/invoice/InvoiceLineService.java
+++ b/src/main/java/org/folio/services/invoice/InvoiceLineService.java
@@ -1,6 +1,7 @@
 package org.folio.services.invoice;
 
 import static java.util.stream.Collectors.toList;
+import static org.folio.invoices.utils.ErrorCodes.ERROR_CREATING_INVOICE_LINE;
 import static org.folio.invoices.utils.ErrorCodes.INVOICE_LINE_NOT_FOUND;
 import static org.folio.invoices.utils.HelperUtils.INVOICE_ID;
 import static org.folio.invoices.utils.ResourcePathResolver.INVOICE_LINES;
@@ -12,12 +13,17 @@ import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.json.JsonObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.folio.invoices.rest.exceptions.HttpException;
 import org.folio.invoices.utils.HelperUtils;
 import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.core.RestClient;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.core.models.RequestEntry;
+import org.folio.rest.jaxrs.model.Error;
 import org.folio.rest.jaxrs.model.Invoice;
 import org.folio.rest.jaxrs.model.InvoiceLine;
 import org.folio.rest.jaxrs.model.InvoiceLineCollection;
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/invoices/utils/HelperUtils.java
```diff
@@ -168,6 +172,30 @@ public static <T> Future<List<T>> collectResultsOnSuccess(List<Future<T>> future
       .map(CompositeFuture::list);
   }
 
+  public static <I, O> Future<List<O>> executeWithSemaphores(Collection<I> collection,
+      FunctionReturningFuture<I, O> f, RequestContext requestContext) {
```
SerhiiNosko, line 176: thanks for introducing this util, we will do the same in mod-orders

SerhiiNosko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/services/invoice/InvoiceLineService.java
```diff
@@ -75,12 +80,19 @@ public Future<List<InvoiceLine>> getInvoiceLinesRelatedForOrder(List<String> ord
         .filter(invoiceLine -> orderPoLineIds.contains(invoiceLine.getPoLineId())).collect(toList()));
   }
 
-  public Future<Void> persistInvoiceLines(List<InvoiceLine> lines,  RequestContext requestContext) {
-    var futures = lines.stream()
-      .map(invoiceLine -> persistInvoiceLine(invoiceLine, requestContext))
-      .collect(Collectors.toList());
-    return GenericCompositeFuture.join(futures).mapEmpty();
+  public Future<List<InvoiceLine>> createInvoiceLines(List<InvoiceLine> invoiceLines,  RequestContext requestContext) {
```
RuslanLavrov, line 83: Hello @damien-git, could you say please can we return invoiceLines creation result (something like InvoiceHelper.createInvoiceAndLines() method) with errors for particular invoice lines in case POST requests fail for some of them? So that we could build such errors for those invoice lines in the CreateInvoiceEventHandler (by INVOICE_LINES_ERRORS_KEY). It will allow to show error only for invoice lines that could not be created, but not for lines that were created.

damien-git: I think an error at this point (when `InvoiceLineService.createInvoiceLines()` is called) means that there is an error for all lines, because other related operations for all lines are not done yet. For instance, the invoice-order relations, the invoice po numbers and the invoice update. Even if an invoice line record is created, the whole operation to create the invoice line (which includes updating the invoice) is not finished. I don't think there is a way around, although I guess we could add code that deletes created lines (if any) when the code fails after the start of the line record creations. Note that this would be an internal error, so it should not happen in normal circumstances.

PR #427, opened 2023-07-25
MODINVOICE-490: Reflection access to private fields fails in Java 17

julianladisch:
https://issues.folio.org/browse/MODINVOICE-490

Replace reflection as suggested in
https://wiki.folio.org/display/TC/JDK+17+and+Java+17#JDK17andJava17-Reflection and remove --add-opens in pom.xml


grabsefx (approved):

NikitaSedyx:
@SinghAdes initial PR to upgrade 17 version has been reverted to unblock testing activities, please pick up these changes and upgrade java one more time, but with proper testing

SinghAdes (approved):

SerhiiNosko (approved):

PR #470, opened 2024-02-08
[MODINVOICE-302] - Improve backend error reporting by including cause in error message systematically

azizbekxm:
## Purpose

https://folio-org.atlassian.net/browse/MODINVOICE-302
## Approach

improved error messages with more detail
## Pre-Merge Checklist:

  - [ ] Duplications on new code are 3% or less


src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -109,7 +113,10 @@ private Future<Void> updateHolderWithExportConfig(BatchVoucherUploadHolder uploa
           uploadHolder.setExportConfig(exportConfigs.getExportConfigs().get(0));
           return null;
         }
-        throw new HttpException(404, "Batch export configuration was not found");
+        var param = new Parameter().withKey("query").withValue(query);
+        var error = new Error().withMessage("Batch export configuration was not found").withParameters(List.of(param));
```
azizbekxm, line 117: question: is it required to introduce error code or is it okay to skip in this case

src/main/java/org/folio/services/voucher/BatchVoucherGenerateService.java
```diff
@@ -73,7 +74,10 @@ public Future<BatchVoucher> buildBatchVoucherObject(BatchVoucherExport batchVouc
     return voucherService.getVouchers(voucherCQL, 0, Integer.MAX_VALUE, requestContext)
       .compose(vouchers -> {
         if (vouchers.getVouchers().isEmpty()) {
-          throw new HttpException(404, "Vouchers for batch voucher export were not found");
+          var param = new Parameter().withKey("voucherCQL").withValue(voucherCQL);
+          var error = new Error().withMessage("Vouchers for batch voucher export were not found").withParameters(List.of(param));
```
azizbekxm, line 78: There is three different message that are similar thinking to skip to introduce error code

damien-git:
@azizbekxm The issue I had in mind when I created that ticket is that users don't see the cause of the error in the UI message when it happened in a called module. For instance mod-invoice calls mod-finance which calls mod-finance-storage, where an error happens. I think the user should be able to see the error message from mod-finance-storage. Since error parameters are currently not displayed to the user, and since the calls can have a large depth, I think we need to include the original error message in the new error message when building an HttpException. For instance, replace:
```
throw new HttpException(500, ORDER_INVOICE_RELATION_CREATE_FAILED.toErrors(), throwable.getMessage());
```
by something like:
```
String message = String.format(ORDER_INVOICE_RELATION_CREATE_FAILED.getDescription(), throwable.getMessage());
Error error = new Error().withCode(ORDER_INVOICE_RELATION_CREATE_FAILED.getCode()).withMessage(message);
throw new HttpException(500, error);
```
and add a ": %s" to the end of ORDER_INVOICE_RELATION_CREATE_FAILED.
Actually I am not sure the message will be displayed to the user because there is a code and it gets localized. But that might be something to fix in the UI.

azizbekxm:
> is a code and it gets lo

Thanks to clarification, what about the situation where there is no any throwable error just condition like your pr https://github.com/folio-org/mod-finance-storage/pull/340

src/main/java/org/folio/rest/impl/FundDistributionsValidationHelper.java
```diff
@@ -43,7 +43,10 @@ public Future<Void> validateFundDistributions(ValidateFundDistributionsRequest r
         List<FundDistribution> fundDistributionList = request.getFundDistribution();
         if (CollectionUtils.isNotEmpty(request.getAdjustments())) {
           if (validator.isAdjustmentIdsNotUnique(request.getAdjustments())) {
-            throw new HttpException(400, ADJUSTMENT_IDS_NOT_UNIQUE);
+            var parameter = new Parameter().withKey("adjustments").withValue(request.getAdjustments().toString());
+            var error = ADJUSTMENT_IDS_NOT_UNIQUE.toError().withParameters(List.of(parameter));
+            logger.error(JsonObject.mapFrom(error).encodePrettily());
```
SerhiiNosko, line 48: better to add some message before encoding error, in this case it will be easier to find error in the logs by matching with  phrase from source code

SerhiiNosko: and please do it for other places

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -365,7 +364,10 @@ public Future<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine) {
       .compose(v -> createInvoiceLine(ilProcessing, requestContext))
       .compose(v -> orderService.createInvoiceOrderRelation(ilProcessing.getInvoiceLine(), buildRequestContext())
         .recover(throwable -> {
-          throw new HttpException(500, ORDER_INVOICE_RELATION_CREATE_FAILED.toError());
+          String message = String.format(ORDER_INVOICE_RELATION_CREATE_FAILED.getDescription(), throwable.getMessage());
+          var error = ORDER_INVOICE_RELATION_CREATE_FAILED.toError().withMessage(message);
+          logger.error("Failed to create invoice order relation", throwable);
+          throw new HttpException(500, error);
```
SerhiiNosko, line 370: can you provide screen how it will be shown in browsers Network response, it's basically can be done for any error code, not exactly this one

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -374,7 +376,11 @@ public Future<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine) {
 
   private void checkIfInvoiceLineCreationAllowed(Invoice invoice) {
     if (isPostApproval(invoice)) {
-      throw new HttpException(500, PROHIBITED_INVOICE_LINE_CREATION);
+      var param = new Parameter().withKey("invoiceStatus").withValue(invoice.getStatus().toString());
+      var error = PROHIBITED_INVOICE_LINE_CREATION.toError().withParameters(List.of(param));
+      if (logger.isErrorEnabled())
```
SerhiiNosko, line 381: this if is not necessary

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -449,7 +455,7 @@ private Future<List<InvoiceLine>> applyProratedAdjustments(InvoiceLine invoiceLi
       return adjustmentsService.applyProratedAdjustments(allLines, invoice)
         .stream()
         .filter(line -> !line.equals(invoiceLine))
-        .collect(toList());
+        .toList();
```
SerhiiNosko, line 458: when making such changes please check that returned list is unmodifiable list, we need to make sure that the code not modify this collection

src/main/java/org/folio/services/invoice/InvoiceCancelService.java
```diff
@@ -167,25 +166,25 @@ private Future<Void> unreleaseEncumbrances(List<InvoiceLine> invoiceLines, Invoi
...
       .compose(poLines -> unreleaseEncumbrancesForPoLines(poLines, invoiceFromStorage, requestContext))
       .recover(t -> {
         Throwable cause = requireNonNullElse(t.getCause(), t);
-        List<Parameter> parameters = Collections.singletonList(
-          new Parameter().withKey("cause").withValue(cause.toString()));
-        Error error = ERROR_UNRELEASING_ENCUMBRANCES.toError().withParameters(parameters);
+        var param = new Parameter().withKey("cause").withValue(cause.toString());
+        String message = String.format(ERROR_UNRELEASING_ENCUMBRANCES.getDescription(), t.getMessage());
+        var error = ERROR_UNRELEASING_ENCUMBRANCES.toError().withMessage(message).withParameters(List.of(param));
         logger.error(error.getMessage(), cause);
```
SerhiiNosko, line 188: We already have cause in this log, so better to add some message that we can use for quickly search by, because there can be lot of defferent messages
```suggestion
        logger.error("Failed to unrelease encumbrance for po lines", cause);
```

azizbekxm: Thanks, Changed error message

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
azizbekxm: Thanks to remind. reverted some of them just in case

src/main/java/org/folio/rest/impl/FundDistributionsValidationHelper.java
azizbekxm: changed, error messages to meaningful one

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -471,10 +468,9 @@ private void verifyTransitionOnPaidStatus(Invoice invoiceFromStorage, Invoice in
     // Once an invoice is Paid, it should no longer transition to other statuses, except Cancelled.
     if (invoiceFromStorage.getStatus() == Invoice.Status.PAID && invoice.getStatus() != Invoice.Status.CANCELLED &&
       invoice.getStatus() != invoiceFromStorage.getStatus()) {
-      List<Parameter> parameters = Collections.singletonList(new Parameter().withKey("invoiceId")
-        .withValue(invoice.getId()));
-      Error error = INVALID_INVOICE_TRANSITION_ON_PAID_STATUS.toError()
-        .withParameters(parameters);
+      var parameter = new Parameter().withKey("invoiceId").withValue(invoice.getId());
+      Error error = INVALID_INVOICE_TRANSITION_ON_PAID_STATUS.toError().withParameters(List.of(parameter));
+      logger.error("verifyTransitionOnPaidStatus:: Invalid invoice transition on paid status. '{}'", parameter);
```
SerhiiNosko, line 473: I don't see that toString() is overridden for Parameter class, so we will not have some meaningful messages.
Also this looks a bit overengineered, it would be enough just to add invoice Id to the log.
Also please change in another places as well

src/main/java/org/folio/rest/impl/InvoiceHelper.java
```diff
@@ -522,10 +522,10 @@ private void validateBeforeApproval(Organization organization, Invoice invoice,
     if (organization == null) {
       throw new HttpException(404, ORG_NOT_FOUND);
     }
-    if (!organization.getIsVendor()) {
+    if (Boolean.FALSE.equals(organization.getIsVendor())) {
       var param = new Parameter().withKey("organizationId").withValue(organization.getId());
       var error = ORG_IS_NOT_VENDOR.toError().withParameters(List.of(param));
-      logger.error(JsonObject.mapFrom(error).encodePrettily());
+      logger.error("validateBeforeApproval:: Organization is not vendor. '{}'", param);
```
SerhiiNosko, line 528: the same here, just adding organizationId would be more simpler and readable

azizbekxm: Thanks, improved log.errors

src/main/java/org/folio/rest/impl/InvoiceHelper.java
azizbekxm: Thanks, using only invoice getId

src/main/java/org/folio/invoices/utils/ErrorCodes.java
```diff
@@ -21,7 +21,7 @@ public enum ErrorCodes {
   INCORRECT_FUND_DISTRIBUTION_TOTAL("incorrectFundDistributionTotal","Fund distribution total must add to 100% or totalPrice"),
   CANNOT_MIX_TYPES_FOR_ZERO_PRICE("cannotMixTypesForZeroPrice", "Fund distribution types cannot be mixed for a price of 0"),
   ADJUSTMENT_FUND_DISTRIBUTIONS_SUMMARY_MISMATCH("adjustmentFundDistributionsSummaryMismatch", "Fund distributions summary should be 100 % or equal to subtotal for every non-prorated adjustment"),
-  VOUCHER_UPDATE_FAILURE("voucherUpdateFailure", "Voucher record cannot be updated"),
+  VOUCHER_UPDATE_FAILURE("voucherUpdateFailure", "Voucher record cannot be updated: %s"),
```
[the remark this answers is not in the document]
SerhiiNosko: CC: @damien-git

damien-git:
> Thanks to clarification, what about the situation where there is no any throwable error just condition like your pr [folio-org/mod-finance-storage#340](https://github.com/folio-org/mod-finance-storage/pull/340)

Handling errors created in mod-invoice is not in the scope of this ticket, only handling errors that were created in a different module. Of course it doesn't hurt if you generally improve error handling.

src/main/java/org/folio/invoices/utils/ErrorCodes.java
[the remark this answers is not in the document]
damien-git: In general it is good to use parameter. The problem with them is that there is only one level of parameters, so if you want to report an error that happened in another module that is using parameters and you want to put that error in a parameter, you have to also add the parameters of that error to the new parameter value. I thought it was simpler to create long messages including the cause because that works with no limit in depth. But that is only for errors happening in other modules that are caught in this module.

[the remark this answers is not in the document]
azizbekxm: Thanks to clarification guys, I moved String message to parameters as Serhii suggested

src/main/java/org/folio/rest/impl/FundDistributionsValidationHelper.java
```diff
@@ -43,7 +42,10 @@ public Future<Void> validateFundDistributions(ValidateFundDistributionsRequest r
         List<FundDistribution> fundDistributionList = request.getFundDistribution();
         if (CollectionUtils.isNotEmpty(request.getAdjustments())) {
           if (validator.isAdjustmentIdsNotUnique(request.getAdjustments())) {
-            throw new HttpException(400, ADJUSTMENT_IDS_NOT_UNIQUE);
+            var parameter = new Parameter().withKey("adjustments").withValue(request.getAdjustments().toString());
```
SerhiiNosko, line 45: getAdjustments has not toString() implementation, if you want to include it as param you need to use Json encode

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -365,7 +364,10 @@ public Future<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine) {
       .compose(v -> createInvoiceLine(ilProcessing, requestContext))
       .compose(v -> orderService.createInvoiceOrderRelation(ilProcessing.getInvoiceLine(), buildRequestContext())
         .recover(throwable -> {
-          throw new HttpException(500, ORDER_INVOICE_RELATION_CREATE_FAILED.toError());
+          var errorParam = new Parameter().withKey("errorMessage").withValue(throwable.getMessage());
+          var error = ORDER_INVOICE_RELATION_CREATE_FAILED.toError().withParameters(List.of(errorParam));
+          logger.error("Failed to create invoice order relation", throwable);
```
SerhiiNosko, line 369: lets add invoiceLine.getId() to this log

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -374,7 +376,10 @@ public Future<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine) {
 
   private void checkIfInvoiceLineCreationAllowed(Invoice invoice) {
     if (isPostApproval(invoice)) {
-      throw new HttpException(500, PROHIBITED_INVOICE_LINE_CREATION);
+      var param = new Parameter().withKey("invoiceStatus").withValue(invoice.getStatus().toString());
+      var error = PROHIBITED_INVOICE_LINE_CREATION.toError().withParameters(List.of(param));
+      logger.error("checkIfInvoiceLineCreationAllowed:: Prohibited invoice line creation: invoiceStatus={}", invoice.getStatus().toString());
```
SerhiiNosko, line 381: lets add invouiceId to this log

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
SerhiiNosko: if we already introduced new error code - better to use it in all places to be consistent

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -364,9 +364,10 @@ public Future<InvoiceLine> createInvoiceLine(InvoiceLine invoiceLine) {
       .compose(v -> createInvoiceLine(ilProcessing, requestContext))
       .compose(v -> orderService.createInvoiceOrderRelation(ilProcessing.getInvoiceLine(), buildRequestContext())
         .recover(throwable -> {
+          var param = new Parameter().withKey("invoiceLineId").withValue(invoiceLine.getId());
           var errorParam = new Parameter().withKey("errorMessage").withValue(throwable.getMessage());
-          var error = ORDER_INVOICE_RELATION_CREATE_FAILED.toError().withParameters(List.of(errorParam));
-          logger.error("Failed to create invoice order relation", throwable);
+          var error = ORDER_INVOICE_RELATION_CREATE_FAILED.toError().withParameters(List.of(param, errorParam));
+          logger.error("Failed to create invoice '{}' order relation", invoiceLine.getId(), throwable);
```
SerhiiNosko, line 370: Failed to create invoice line '{}' order relation

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -476,8 +483,11 @@ private Future<Void> updateInvoiceAndAffectedLines(ILProcessing ilProcessing, Li
     return persistInvoiceLines(invoice, lines, requestContext)
       .compose(v -> updateInvoice(ilProcessing, requestContext))
       .recover(t -> {
-        logger.error("Failed to update the invoice and other lines", t);
-        throw new HttpException(500, FAILED_TO_UPDATE_INVOICE_AND_OTHER_LINES.toError());
+        var param = new Parameter().withKey("invoiceId").withValue(invoice.getId());
+        var errorParam = new Parameter().withKey("errorMessage").withValue(t.getMessage());
+        var error = FAILED_TO_UPDATE_INVOICE_AND_OTHER_LINES.toError().withParameters(List.of(param, errorParam));
+        logger.error("Failed to update the invoice '{}' and other lines", invoice.getId(), t);
+        throw new HttpException(500, error);
```
SerhiiNosko, line 490: what do you think to introduce new method like this?

void throwHttpException(int httpCode, ErrorCode error, List<Parameter> params)

SerhiiNosko: and another method:

void throwHttpExceptionWithDetails(int httpCode, ErrorCode error, List params, Throwable t) - that will automatically add errorMessage param

src/main/java/org/folio/services/voucher/BatchVoucherService.java
```diff
@@ -48,7 +47,7 @@ public String convertBatchVoucher(BatchVoucher batchVoucher, String contentType)
       try {
         return xmlConverter.marshal(BatchVoucherType.class, xmlBatchVoucher, null,true);
       } catch (XMLStreamException e) {
-        throw new HttpException(400, MARSHAL_ERROR_MSG);
```
SerhiiNosko, line 51: also error log is needed here

src/main/java/org/folio/utils/UserPermissionsUtil.java
```diff
@@ -24,19 +24,22 @@
...
   private UserPermissionsUtil() {
-
   }
 
   public static void verifyUserHasAssignPermission(List<String> acqUnitIds, Map<String, String> okapiHeaders) {
     if (CollectionUtils.isNotEmpty(acqUnitIds) && isUserDoesNotHaveDesiredPermission(ASSIGN, okapiHeaders)) {
-      throw new HttpException(HttpStatus.HTTP_FORBIDDEN.toInt(), USER_HAS_NO_ACQ_PERMISSIONS);
+      var param1 = new Parameter().withKey("Assign").withValue(ASSIGN.getPermission());
+      var param2 = new Parameter().withKey("User permission").withValue(okapiHeaders.get(OKAPI_HEADER_PERMISSIONS));
+      var param3 = new Parameter().withKey("acqUnitIds").withValue(acqUnitIds.toString());
```
SerhiiNosko, line 41: I think these params will not help us, we can revert this file

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
azizbekxm: 1. I also think about the idea, `var error..` is require for log in some places. Since JsonObject.mapFromObject(error) is not used in logs, we can use  `void throwHttpException(int httpCode, ErrorCode error, List params) ` this method is exits. I agree with it will be easier to find correct part with search
2.  void throwHttpExceptionWithDetails(int httpCode, ErrorCode error, List params, Throwable t) - that will automatically add errorMessage param I saw some problem with implementation of method, some unsupported exception while creating custom error message. I think it is okay to use first approach

src/main/java/org/folio/services/voucher/BatchVoucherService.java
azizbekxm: added

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
azizbekxm: thanks, improved message

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
```diff
@@ -476,8 +481,10 @@ private Future<Void> updateInvoiceAndAffectedLines(ILProcessing ilProcessing, Li
     return persistInvoiceLines(invoice, lines, requestContext)
       .compose(v -> updateInvoice(ilProcessing, requestContext))
       .recover(t -> {
+        var errorParam = new Parameter().withKey("errorMessage").withValue(t.getMessage());
+        var error = FAILED_TO_UPDATE_INVOICE_AND_OTHER_LINES.toError().withParameters(List.of(errorParam));
         logger.error("Failed to update the invoice and other lines", t);
```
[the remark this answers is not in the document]
azizbekxm: added identifier, thanks

src/main/java/org/folio/utils/UserPermissionsUtil.java
azizbekxm: Thanks, revereted this file

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
azizbekxm: This style has been changed so based on new approach, when api request this error message can be seen
image

src/main/java/org/folio/rest/impl/InvoiceLineHelper.java
azizbekxm: used void throwHttpException(int httpCode, ErrorCode error, List params)

damien-git:
There is a problem with this kind of implementation:
```
        .recover(throwable -> {
          var param = new Parameter().withKey("invoiceLineId").withValue(invoiceLine.getId());
          var errorParam = new Parameter().withKey("errorMessage").withValue(throwable.getMessage());
          logger.error("Failed to create invoice line '{}' order relation", invoiceLine.getId(), throwable);
          throw new HttpException(500, ORDER_INVOICE_RELATION_CREATE_FAILED, List.of(param, errorParam));
        })
```
Think about an error happening in the called module (which might be caused by another module), and assume another module is calling this one.
This implementation sends the cause message `throwable.getMessage()` in a parameter, but it does not include the error parameters in the resulting error. If the module that is called here does the same thing as here (create a new error and add the error cause in a parameter), the initial cause will not be visible at all in the calling module.

azizbekxm:
this is how error is being shown from other module fails
image
image
image

damien-git:
> this is how error is being shown from other module fails
Good. Could we prettify the thrown message so it can be displayed to end users (with a readable format instead of JSON) ? Any code to do that could be useful in other modules too, so maybe put it in HttpException with a new constructor ?

SerhiiNosko (approved):

SerhiiNosko:
> HttpException

We have localizations for each error code already.
As for me having separate param 'cause' looks better, that adding it to message wiht : separator, because we saw that can see all error params for e.getMessage() statement, but if for you @damien-git the existing option is better - we can remain as is.

damien-git:
> As for me having separate param 'cause' looks better, that adding it to message wiht : separator,

We could do that, but then we need to make sure the cause parameter includes the error parameters. And we need to make sure the parameters are displayed to the end user.

damien-git:
> @damien-git do you agree with this approach, because now is almost the end of the sprint and we need to complete this story?

Sounds good.

PR #471, opened 2024-02-22
MODINVOICE-523 - After resaving the configurations with empty credentials, an NPE occurs during manual export

azizbekxm:
## Purpose

https://folio-org.atlassian.net/browse/MODINVOICE-523
https://folio-org.atlassian.net/browse/MODINVOICE-522

## Approach
Root cause for two ticket is that this place https://github.com/folio-org/mod-invoice/blob/5a511e95da8b4ae4d4d3c4b8145f85c2842586a6/src/main/java/org/folio/services/ftp/FtpUploadService.java#L50
When creating URI from null value it is throwing error. 
image

image
Introducing specific errors for specific cases such as URL is empty, Credential is empty

Added feature to remove empty credentials 
image
image
image


src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -127,7 +129,12 @@ private void updateHolderWithFileFormat(BatchVoucherUploadHolder uploadHolder) {
 
   private Future<Void> updateHolderWithCredentials(BatchVoucherUploadHolder uploadHolder) {
     return batchVoucherExportConfigService.getExportConfigCredentials(uploadHolder.getExportConfig().getId(), requestContext)
-      .onSuccess(uploadHolder::setCredentials)
+      .onSuccess(credentials -> {
```
SerhiiNosko, line 132: onSuccess is invoked when future already executed, so it can be situation when the caller use result of this feature as succesful result but after some time onSuccess will throw an exception, need to use compose or map here

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -150,7 +157,7 @@ private Future<Void> updateHolderWithBatchVoucher(BatchVoucherUploadHolder uploa
   private Future<Void> failUploadUpdate(BatchVoucherExport bvExport, Throwable t) {
     if (bvExport != null) {
       if (!CREDENTIALS_NOT_FOUND.equals(t.getMessage())
-                      && !(t instanceof URISyntaxException)) {
+                      && !(t instanceof URISyntaxException || URL_NOT_FOUND_FOR_FTP.equals(t.getMessage()))) {
```
SerhiiNosko, line 160: need in somehow to simplify this code, propose to delete this if at all

src/main/java/org/folio/services/ftp/FtpUploadService.java
```diff
@@ -26,14 +27,17 @@
 public class FtpUploadService implements FileExchangeService {
 
   private static final Logger logger = LogManager.getLogger(FtpUploadService.class);
-
   private static final String DEFAULT_WORKING_DIR = "/files/invoices";
+  public static final String URL_NOT_FOUND_FOR_FTP = "URL for FTP upload was not found";
```
SerhiiNosko, line 31: need to standartize to the same format, here is URL and in the similar error URI is used

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
azizbekxm: Thanks, Removed the code because of being not valid statement

src/main/java/org/folio/services/ftp/FtpUploadService.java
azizbekxm: Thanks, changed

src/test/java/org/folio/services/ftp/FtpUploadServiceTest.java
```diff
@@ -190,4 +194,27 @@ public void testFailedUpload(VertxTestContext vertxTestContext) throws URISyntax
     vertxTestContext.assertFailure(future)
       .onComplete(result -> vertxTestContext.completeNow());
   }
+
```
SerhiiNosko, line 197: please add test when credentials.username or credentials.password is empty

azizbekxm: yes, working on it...

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -127,7 +128,13 @@ private void updateHolderWithFileFormat(BatchVoucherUploadHolder uploadHolder) {
...
-      .onSuccess(uploadHolder::setCredentials)
+      .compose(credentials -> {
+        if (StringUtils.isBlank(credentials.getUsername()) || StringUtils.isBlank(credentials.getPassword())) {
+          throw new HttpException(404, CREDENTIALS_NOT_FOUND);
+        }
+        uploadHolder.setCredentials(credentials);
+        return Future.succeededFuture(null);
+      })
       .recover(t -> {
         throw new HttpException(404, CREDENTIALS_NOT_FOUND);
```
SerhiiNosko, line 139: I think we can improve more here, lets imagine that we throw an exception with another error code in compose section, in this case recover will always rewrite original exception.
As solution we can add .recover after .getExportConfigCredentials and at the end use compose where checking credentials

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -149,8 +158,8 @@ private Future<Void> updateHolderWithBatchVoucher(BatchVoucherUploadHolder uploa
 
   private Future<Void> failUploadUpdate(BatchVoucherExport bvExport, Throwable t) {
     if (bvExport != null) {
-      if (!CREDENTIALS_NOT_FOUND.equals(t.getMessage())
-                      && !(t instanceof URISyntaxException)) {
+      if (Boolean.FALSE.equals(URL_NOT_FOUND_FOR_FTP.equals(t.getMessage()) || URI_SYNTAX_ERROR.equals(t.getMessage())
+        || CREDENTIALS_NOT_FOUND.equals(t.getMessage()))) {
```
azizbekxm, line 162: After talk with @Abdulkhakimov, decied that the cases connected with not found exception shouildn't change the status.
However, we can discuss this part

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
azizbekxm: This validation was removed, because empty credentials are no longer saved to db

azizbekxm: https://github.com/folio-org/mod-invoice/pull/471/commits/27b3170dc6869c4b481ef0205ca0d22a2328a293#diff-bf1e298570cf67f2736194d7d458d26e8cb6c2933ba1d9b9e352f520f920bb05L58-R73

src/main/java/org/folio/rest/impl/BatchVoucherExportConfigHelper.java
```diff
@@ -65,7 +68,9 @@ public Future<Void> putExportConfig(ExportConfig exportConfig) {
 
   public Future<Void> putExportConfigCredentials(String id, Credentials credentials) {
     String path = String.format(resourcesPath(BATCH_VOUCHER_EXPORT_CONFIGS_CREDENTIALS), id);
-    return restClient.put(path, credentials, buildRequestContext());
+    return StringUtils.isBlank(credentials.getUsername()) || StringUtils.isBlank(credentials.getPassword()) ?
```
SerhiiNosko, line 71: we need some logs in these 2 places in this class, because this logic is not so obvious

SerhiiNosko (approved):

src/main/java/org/folio/rest/impl/UploadBatchVoucherExportHelper.java
```diff
@@ -149,8 +151,8 @@ private Future<Void> updateHolderWithBatchVoucher(BatchVoucherUploadHolder uploa
 
   private Future<Void> failUploadUpdate(BatchVoucherExport bvExport, Throwable t) {
     if (bvExport != null) {
-      if (!CREDENTIALS_NOT_FOUND.equals(t.getMessage())
-                      && !(t instanceof URISyntaxException)) {
+      if (Boolean.FALSE.equals(URL_NOT_FOUND_FOR_FTP.equals(t.getMessage()) || URI_SYNTAX_ERROR.equals(t.getMessage())
```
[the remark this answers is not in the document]
azizbekxm: This part actually like this `Boolean.FALSE.equals(URL_NOT_FOUND_FOR_FTP.equals(t.getMessage()) || URI_SYNTAX_ERROR.equals(t.getMessage()) || CREDENTIALS_NOT_FOUND.equals(t.getMessage()))`
if it will be changed to `if(!(URL_NOT_FOUND_FOR_FTP.equals(t.getMessage()) || URI_SYNTAX_ERROR.equals(t.getMessage()) || CREDENTIALS_NOT_FOUND.equals(t.getMessage()))))
`
I thought it will make code hard to read a little bit, so that is why Boolean.FALSE.equals is being used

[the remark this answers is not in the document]
imerabishvili: Got it, it's already very hard to read for me : )
We have a following semantics:
```
      if (
        !URL_NOT_FOUND_FOR_FTP.equals(message)
        && !URI_SYNTAX_ERROR.equals(message)
        && !CREDENTIALS_NOT_FOUND.equals(message)
      ) {
```
What you make hard to read is set of negations and conjunctions ( !a && !b && !c).
In this case you can have a variable `x = a || b || c` and use `!x`:
```
      boolean isKnownError =
        URL_NOT_FOUND_FOR_FTP.equals(message) || 
        URI_SYNTAX_ERROR.equals(message) ||
        CREDENTIALS_NOT_FOUND.equals(message);
      if (!isKnownError) {
        bvExport.setStatus(BatchVoucherExport.Status.ERROR);
      }
```
This also clarifies that we set error status when exception is not known or whatever variable name says.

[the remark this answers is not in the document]
azizbekxm: Thanks, code was improved

imerabishvili (approved):

PR #483, opened 2024-04-01
Release v5.8.1

Abdulkhakimov:


SerhiiNosko (approved):

azizbekxm (approved):
Very very small suggestion, usually we put tickets in decreasing order
modinvoice-31
modinvoice-21
may be not need to action, just in case

Abdulkhakimov:
> Very very small suggestion, usually we put tickets in decreasing order modinvoice-31 modinvoice-21 may be not need to action, just in case

I see. Thanks. Reordered!

PR #496, opened 2024-05-28
MODINVOICE-548 - Invoices do not display fund codes

RuslanLavrov:
## Purpose
to ensure fund code in imported invoice lines when fund distributions should be populated from the field mappings

## Approach
* set invoice line fund code based on fundId populated during mapping
* add test

## Learning
[MODINVOICE-548](https://issues.folio.org/browse/MODINVOICE-548)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

Maksat-Galymzhan (approved):

src/main/java/org/folio/dataimport/handlers/actions/CreateInvoiceEventHandler.java
```diff
@@ -372,9 +381,45 @@ private List<InvoiceLine> prepareInvoiceLinesToSave(String invoiceId, DataImport
...
+      .flatMap(mappingRule -> mappingRule.getSubfields().get(0).getFields().stream())
+      .filter(mappingRule -> FUND_ID_RULE_NAME.equals(mappingRule.getName()))
+      .map(mappingRule -> ((Map<String, String>) mappingRule.getAcceptedValues()))
+      .findAny()
+      .orElse(Collections.emptyMap());
+  }
+
+  private void populateFundCode(org.folio.rest.jaxrs.model.FundDistribution fundDistribution, Map<String, String> idToFundName) {
+    String fundName = idToFundName.get(fundDistribution.getFundId());
+    fundDistribution.setCode(fundName.substring(fundName.lastIndexOf(LEFT_BRACKET) + 1, fundName.lastIndexOf(RIGHT_BRACKET)));
```
JavokhirAbdullayev, line 420: can we get null here for fundName?

JavokhirAbdullayev (approved):

VRohach (approved):

azizbekxm (approved):

src/main/java/org/folio/dataimport/handlers/actions/CreateInvoiceEventHandler.java
RuslanLavrov: we can not because this method is called when fundId is not null and idToFundName data is retrieved from mapping profile. fundId is populated when a mapping profile contains an appropriate pair of id and fund name that matches to fund name specified in the mapping for "fundId" field, so if "fundId" field is populated then a mapping profile and idToFundName map contains pair of fund id with corresponding fund name

PR #500, opened 2024-06-28
[MODINVOICE-547] - Separate credits from expenditures in mod-invoice

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/MODINVOICE-547
## Approach
Changed calculation 
replace: [remaining amount] = (allocated + netTransfers) * allowableExpenditure - (awaitingPayment + expended)
by: [remaining amount] = (allocated + netTransfers) * allowableExpenditure - (awaitingPayment + expended - credited)

### DIFF
--- a/src/main/java/org/folio/services/validator/FundAvailabilityHolderValidator.java
+++ b/src/main/java/org/folio/services/validator/FundAvailabilityHolderValidator.java
@@ -79,11 +79,12 @@ public class FundAvailabilityHolderValidator implements HolderValidator {
 
   private boolean isRemainingAmountExceed(Budget budget, MonetaryAmount totalExpendedAmount) {
     // [remaining amount we can expend] = (totalFunding * allowableExpenditure) - expended
-    // where expended = awaitingPayment + expenditure
+    // where expended = awaitingPayment + expenditure - credited
     CurrencyUnit currency = totalExpendedAmount.getCurrency();
     Money totalFundings = Money.of(budget.getTotalFunding(), currency);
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/services/validator/FundAvailabilityHolderValidator.java
```diff
@@ -77,17 +76,21 @@ private MonetaryAmount calculateTotalExpendedAmount(List<InvoiceWorkflowDataHold
...
   }
 
+  /**
+   * Method is following these formulas <br>
+   * afterApproveExpended [remaining amount we can expend] = expended - credited + (totalFunding * allowableExpenditure) <br>
+   * expended = awaitingPayment + expenditure
+   */
   private boolean isRemainingAmountExceed(Budget budget, MonetaryAmount totalExpendedAmount) {
-    // [remaining amount we can expend] = (totalFunding * allowableExpenditure) - expended
-    // where expended = awaitingPayment + expenditure
```
azizbekxm, line 82: regarding this changes the logic is not the same with comment, so updated comment based on the code.

azizbekxm:
image

src/main/java/org/folio/services/validator/FundAvailabilityHolderValidator.java
damien-git: This code is very confusing, please create a `awaitingPayment` variable to separate it from expended and rename `totalExpendedAmount` to `additionalAmountToExpend`.

damien-git (approved):

PR #520, opened 2024-12-20
MODINVOICE-568 Use system user during data-import invoice creation

dmytrokrutii:
* [MODINVOICE-568](https://issues.folio.org/browse/MODINVOICE-568) Use system user for handling DI events

To support DI operation in Eureka, we should implement support for performing asynchronous operations on behalf of a system user. 

## Approach
Add system user to handle data-import events with null or empty token

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1113,10 +1113,13 @@
         "invoice-storage.invoices.item.get",
         "invoice-storage.invoices.item.put",
         "acquisitions-units-storage.units.collection.get",
-        "acquisitions-units.memberships.collection.get",
-        "acquisitions-units.units.collection.get",
+        "acquisitions-units-storage.units.collection.get",
+        "acquisitions-units-storage.memberships.collection.get",
         "orders.po-lines.collection.get",
-        "orders.po-lines.item.get"
+        "orders.po-lines.item.get",
+        "orders-storage.order-invoice-relationships.collection.get",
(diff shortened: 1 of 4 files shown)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -1097,6 +1097,29 @@
       "description": "Ability to manage acquisition unit and update invoice"
     }
   ],
+  "metadata": {
+    "user": {
+      "type": "system",
+      "permissions": [
```
RuslanLavrov, line 1103: please add these permissions as well:
"orders-storage.order-invoice-relationships.collection.get"
"orders-storage.order-invoice-relationships.item.post"
"acquisitions-units-storage.memberships.collection.get"

Also, looks like endpoints usage was changed and these permissions have started to be used: 
"acquisitions-units-storage.units.collection.get", "acquisitions-units-storage.memberships.collection.get".
And these are not required:
"acquisitions-units.memberships.collection.get", "acquisitions-units.units.collection.get".

VRohach: @RuslanLavrov  Done! can you double-check please?

RuslanLavrov: thanks!

RuslanLavrov (approved):

PR #533, opened 2025-04-09
[MODINVOICE-579]. Switch to using the Finance Exchange Rate Api

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODINVOICE-579>

## Approach

- Add `CacheableExchangeRateService` that will be called before `CentralExchangeRateProvider` is initialized
- `CentralExchangeRateProvider` will handle both dynamic API-based and custom exchange rates
- Disable all built-in conversion providers (i.e. local implementation of ECB) to rely fully on **mod-finance**
- Update and clean test classes, remove some deprecated methods and add more unit tests


src/main/java/org/folio/services/voucher/VoucherCommandService.java
```diff
@@ -51,35 +48,39 @@ public Future<Voucher> buildNewVoucher(Invoice invoice, RequestContext requestCo
...
-      ExchangeRateProvider exchangeRateProvider = exchangeRateProviderResolver.resolve(conversionQuery, requestContext);
-      ExchangeRate exchangeRate = exchangeRateProvider.getExchangeRate(conversionQuery);
-      invoice.setExchangeRate(exchangeRate.getFactor().doubleValue());
-      voucher.setExchangeRate(exchangeRate.getFactor().doubleValue());
-      event.complete(voucher);
-    });
+    if (Objects.nonNull(invoice.getExchangeRate())) {
+      voucher.setExchangeRate(invoice.getExchangeRate());
+      return Future.succeededFuture(voucher);
+    }
```
SerhiiNosko, line 77: maybe we do not need this check because later in method we set voucher exchange rate already

BKadirkhodjaev: > maybe we do not need this check because later in method we set voucher exchange rate already

Yeah, I forgot to remove it, will do now

SerhiiNosko (approved):

src/main/java/org/folio/invoices/utils/ResourcePathResolver.java
```diff
@@ -76,6 +77,7 @@ private ResourcePathResolver() {
     apis.put(FINANCE_EXCHANGE_RATE, "/finance/exchange-rate");
     apis.put(TENANT_CONFIGURATION_ENTRIES, "/configurations/entries");
     apis.put(FISCAL_YEARS, "/finance/fiscal-years");
+    apis.put(EXCHANGE_RATE, "/finance/exchange-rate");
```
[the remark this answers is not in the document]
SerhiiNosko: its already presented in MD

azizbekxm (approved):

PR #564, opened 2025-09-03
[MODINVOICE-608]. Negative encumbrances on orders with paid invoices - LOC

BKadirkhodjaev:
### **Purpose**

- <https://folio-org.atlassian.net/browse/MODINVOICE-608>

### **Approach**

- Add code to prevent unnecessary unreleasing of encumbrances
- Associate Karate PR: <https://github.com/folio-org/folio-integration-tests/pull/2009>

### DIFF
--- a/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
+++ b/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
@@ -30,6 +30,7 @@ import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.Invoice;
 import org.folio.rest.jaxrs.model.InvoiceLine;
 import org.folio.rest.jaxrs.model.Parameter;
+import org.folio.services.finance.EncumbranceUtils;
 import org.folio.services.finance.transaction.BaseTransactionService;
 import org.folio.services.finance.transaction.EncumbranceService;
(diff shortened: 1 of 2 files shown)


SerhiiNosko (approved):

src/main/java/org/folio/services/invoice/InvoiceCancelService.java
```diff
@@ -221,6 +221,10 @@ private Future<Void> unreleaseEncumbrancesForPoLines(List<PoLine> poLines, Invoi
     return encumbranceService.getEncumbrancesByPoLineIds(poLineIds, fiscalYearId, requestContext)
       .map(transactions -> transactions.stream()
         .filter(tr -> RELEASED.equals(tr.getEncumbrance().getStatus()))
+        // only unrelease encumbrances with expended + credited + awaiting payment = 0
+        .filter(tr -> tr.getEncumbrance().getAmountExpended() == 0
+          && tr.getEncumbrance().getAmountCredited() == 0
+          && tr.getEncumbrance().getAmountAwaitingPayment() == 0)
```
SerhiiNosko, line 227: I so we have the same logic in mod-orders, probably safer would be to create the same EncumbranceUtils even if we have usage only in 1 place here, in some future we need to think to move this Encumbrance utils to some shared library, maybe create additional subproject with jar executable in mod-finance

PR #617, opened 2026-04-24
[MODORDERS-1439] Cancel invoice: update po line statuses before unreleasing encumbrances

damien-git:
## Purpose
[MODORDERS-1439](https://folio-org.atlassian.net/browse/MODORDERS-1439) - Cancelling an invoice does not update order correctly

## Approach
- When cancelling an invoice, update po line statuses before unreleasing encumbrances
- Wait for the order status to have a chance to update async after changing the po line statuses

## Related PRs
- [mod-orders](https://github.com/folio-org/mod-orders/pull/1291)
- [folio-integration-tests](https://github.com/folio-org/folio-integration-tests/pull/2295)

---

### **Pre-Review Checklist**

- [ ] **Self-reviewed Code** — Reviewed code for issues, unnecessary parts, and overall quality.
- [ ] **Change Notes** — NEWS.md updated with clear description and issue key.
- [ ] **Testing** — Confirmed changes were tested locally or on dev environment.
- [ ] **Logging**: Confirmed that logging is appropriately handled.
- [ ] **Breaking Changes** — Handled all required actions if changes affect API, DB, or interface versions.
  - [ ] API/schema changes
  - [ ] Interface version updates
  - [ ] DB schema changes / migration scripts
- [ ] **New Properties / Environment Variables** — Updated README.md if new configs were added.

### DIFF
--- a/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
+++ b/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
@@ -108,7 +108,7 @@ public class InvoiceCancelService {
       .compose(v -> cancelVoucher(invoiceId, requestContext))
       .compose(v -> poLinePaymentStatusUpdateService.updatePoLinePaymentStatusToCancelInvoice(invoiceFromStorage,
         lines, poLinePaymentStatus, requestContext))
-      .compose(v -> vertx.timer(50)) // wait for order to have a chance to reopen automatically
+      .compose(v -> vertx.timer(200)) // wait for order to have a chance to reopen automatically
       .compose(v -> updateOrUnreleaseEncumbrances(lines, invoiceFromStorage, requestContext))
       .onSuccess(v -> log.info("cancelInvoice:: Invoice {} cancelled successfully", invoiceId))
       .onFailure(t -> log.error("cancelInvoice:: Failed to cancel invoice {}", invoiceId, t));
--- a/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
+++ b/src/main/java/org/folio/services/invoice/InvoiceCancelService.java
@@ -108,8 +108,11 @@ public class InvoiceCancelService {
       .compose(v -> cancelVoucher(invoiceId, requestContext))
       .compose(v -> poLinePaymentStatusUpdateService.updatePoLinePaymentStatusToCancelInvoice(invoiceFromStorage,
         lines, poLinePaymentStatus, requestContext))
-      .compose(v -> vertx.timer(200)) // wait for order to have a chance to reopen automatically
-      .compose(v -> updateOrUnreleaseEncumbrances(lines, invoiceFromStorage, requestContext))
+      .onSuccess(v -> {
+        // Wait 1s for orders to reopen and unrelease encumbrances asynchronously
+        vertx.timer(1000)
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/services/invoice/InvoiceCancelService.java
```diff
@@ -91,7 +92,7 @@ public Future<Void> cancelInvoice(Invoice invoiceFromStorage, List<InvoiceLine>
                                     RequestContext requestContext) {
     String invoiceId = invoiceFromStorage.getId();
     log.info("cancelInvoice:: Cancelling invoice {}...", invoiceId);
-
+    Vertx vertx = Vertx.vertx();
```
SerhiiNosko: Vertx.vertx() is a factory that creates a fresh Vertx instance with its own event-loop pool, worker pool, file resolver, etc. It's also never closed. Every invoice cancel will leak threads.
Use  requestContext.getContext().owner() instead as in RestClient

src/main/java/org/folio/services/invoice/InvoiceCancelService.java
```diff
@@ -105,9 +106,10 @@ public Future<Void> cancelInvoice(Invoice invoiceFromStorage, List<InvoiceLine>
         return null;
       })
       .compose(v -> cancelVoucher(invoiceId, requestContext))
-      .compose(v -> updateOrUnreleaseEncumbrances(lines, invoiceFromStorage, requestContext))
       .compose(v -> poLinePaymentStatusUpdateService.updatePoLinePaymentStatusToCancelInvoice(invoiceFromStorage,
         lines, poLinePaymentStatus, requestContext))
+      .compose(v -> vertx.timer(50)) // wait for order to have a chance to reopen automatically
```
SerhiiNosko: Ideally to replace fixed-time wait, I think we can increase to 100 ms probably, but not very like this approach.
Options is to use polling if the status is correct before invoking updateOrUnreleaseEncumbrances.
Another option is introduce new endpoint in mod-orders like reEvaluateAffectedOrders and so move this logic to mod-orders

src/main/java/org/folio/services/invoice/InvoiceCancelService.java
damien-git: OK

src/main/java/org/folio/services/invoice/InvoiceCancelService.java
damien-git: I thought about polling, but we don't have a way within mod-invoice to know what the expected status would be for all affected orders (at least, not without duplicating logic from mod-orders, which would be bad).
Moving the logic to mod-orders would be very complex, and would probably require another endpoint or message handler. I don't think it would be a good solution either.
I checked it was working with 25ms, so I think it's relatively safe with 50ms. In the worst case some encumbrances would not be unreleased, which would not be a major problem. It's not perfect, but I think it's a good enough solution given this is an edge case that would not cause data corruption.

Perhaps a better solution would be to handle the order status change in mod-orders without async. It would slow down line processing a bit, but it might help make other workflows safer. But I think we would need to do that in the scope of another ticket if we wanted to.

SerhiiNosko: I think then we should at least write todo and insert jira in that comment ticket when it will be resolved

